Issue #3, opened 2017-09-30
Add support to provide an external ObjectMapper


srinivasvsk:
Firstly,   greatly appreciate your work on simplifying the json configuration for hibernate. 
While using this package with kotlin though, ObjectMapper needs some additional configuration,  would you look at externalizing the objectmapper, so it can be configured from outside or re-used.

Thanks again

vladmihalcea:
Just send me a Pull Request with your proposed changes and I'll review them and let you know if they need any further improvements.

UltimaPhoenix:
I think that this is an important feature, I will send you a Pull Request for this.

vladmihalcea:
@UltimaPhoenix There already [this branch](https://github.com/vladmihalcea/hibernate-types/tree/CustomObjectMapper) where I started the development. I just didn't have time to finish it for all modules. 

You can try it. It works just fine on the branch.

mguilherme:
Are there any news regarding this issue @vladmihalcea ?

Right now I'm providing additional configuration to the ObjectMapper like this:

```Java
	/**
	 * {@link JacksonUtil#OBJECT_MAPPER} configuration.
	 */
	static {
		JacksonUtil.OBJECT_MAPPER.configure( DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false );

//		JacksonUtil.OBJECT_MAPPER.configure( SerializationFeature.INDENT_OUTPUT, true );
		JacksonUtil.OBJECT_MAPPER.configure( SerializationFeature.FAIL_ON_EMPTY_BEANS, false );

		JacksonUtil.OBJECT_MAPPER.enableDefaultTyping( ObjectMapper.DefaultTyping.NON_CONCRETE_AND_ARRAYS, As.PROPERTY );
		JacksonUtil.OBJECT_MAPPER.disable( SerializationFeature.WRITE_DATES_AS_TIMESTAMPS );

		JacksonUtil.OBJECT_MAPPER.registerModules( new JavaTimeModule() );
	}
```

It looks like the `JacksonUtil.OBJECT_MAPPER.enableDefaultTyping( ObjectMapper.DefaultTyping.NON_CONCRETE_AND_ARRAYS, As.PROPERTY );` is not working, I guess If I try to use non concrete types it will fail.

vladmihalcea:
[...]

PR #4, opened 2017-10-23
Find and register Jackson Modules automatically based on classpath.

pedroxs:
Should fix issue #3, but it would still be a good idea to expose the 'ObjectMapper' to the user for further customizations.

@vladmihalcea please let me know if you need anything else to merge / deploy this fix as I would really like to start using it from Maven Repos.

Best regards,
Joaquim Pedro


vladmihalcea:
@pedroxs Thanks. Looks good to me. I'll release the 1.1.0 version with a new `PostgreSQLEnumType` as well. When I have time, I'll see what I could do about the configurable `ObjectMapper`.

vladmihalcea:
I released version 1.1.0 which includes your changes.

PR #15, opened 2017-11-26
Serialization problem when using JsonNodeBinaryType and Hibernate 2nd-level cache

andyflury:
Hi Vlad

Here is the test case that fails. 

Thanks
Andy

### DIFF
--- a/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonNodeTypeDescriptor.java
+++ b/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonNodeTypeDescriptor.java
@@ -1,6 +1,9 @@
 package com.vladmihalcea.hibernate.type.json.internal;
 
 import com.fasterxml.jackson.databind.JsonNode;
+
+import java.io.Serializable;
+
 import org.hibernate.type.descriptor.WrapperOptions;
 import org.hibernate.type.descriptor.java.AbstractTypeDescriptor;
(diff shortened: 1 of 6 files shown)


andyflury:
and here is my suggestion on how to fix the issue

hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonNodeTypeDescriptor.java
```diff
@@ -15,6 +18,17 @@
 
     public JsonNodeTypeDescriptor() {
         super(JsonNode.class, new MutableMutabilityPlan<JsonNode>() {
+
+            @Override
+            public Serializable disassemble(JsonNode value) {
```
vladmihalcea: Looks good. The `MutabilityPlan` is the one who takes care of the marshalling/unmarshalling. That's why these two methods are final in the Hibernate Type.

vladmihalcea:
I'm going to integrate it tomorrow.

vladmihalcea:
Applied upstream, thanks

andyflury:
Thank you so much Vlad!

vladmihalcea:
Thank you for the fix. I'll release it after I integrate the other Pull Request for List/Set/Maps.

vladmihalcea:
This feature is now available if you upgrade to 1.2.0.

andyflury:
great! thank you so much Vlad

PR #22, opened 2018-02-03
Added cause to JacksonUtil exceptions

LiamMorrow:
This helps dramatically when attempting to debug why a model has failed to be deserialized


vladmihalcea:
Thanks. Applied upstream.

PR #26, opened 2018-03-06
Injectable json mapper for type descriptors

lgoldstein:
Great project - however, in order to make it more flexible I believe it should allow the users to provide their own _Jackson_ `ObjectMapper` instance just in case the default is not set-up they way they need it. For anyone else the default may be good enough, but it would help (a lot) if users are able to also do the following:

```java
public static final ObjectMapper MY_SUPER_DUPER_OBJECT_MAPPER = ....some special initialization ...

public class MyJsonBinaryType extends JsonBinaryType {
    public MyJsonNodeBinaryType() {
        super(MY_SUPER_DUPER_OBJECT_MAPPER);
    }
}

@TypeDefs({
    @TypeDef(name = "jsonb", typeClass = MyJsonBinaryType.class),
})
@MappedSuperclass
public class MyBaseEntity {
   ...
}
```

**Note**: I made this change only for _Hibernate_ 5.2 since it is the latest one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,50 @@ hs_err_pid*
 .idea/
 target/
 *.iml
+
+# Eclipse workspace
+.metadata/
+
+# Eclipse Settings directory
+.settings/
+
+# Eclipse Java project classpath
+.classpath
+
+# Eclipse Project file
+.project
+
+# Eclipse PyDev project
+.pydevproject
+
+# Eclipse Checkstyle plugin configuration
+.checkstyle
+
+# Eclipse PMD plugin configuration
+.eclipse-pmd
+.pmd
+
+# Eclipse (R)emote (S)ervers (E)xtensions
+RemoteSystemsTempFiles/
+
+# Eclipse application servers
+Servers/
+
+# Eclipse code recommenders
+.recommenders/
+
+# Eclipse tool builders configurations overrides
+.externalToolBuilders/
+
+# Eclipse Spring Tools Suite
+.springBeans
+
+# Eclipse Java annotation processor (APT)
+.factorypath
+
+# Files open for VI
+*.swp
+
+# IntelliJ workspace
+.idea
+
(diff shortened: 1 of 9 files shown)


vladmihalcea:
Thanks for the PR. I'd like to see some unit tests to see how you are thinking of using it.

I created [this branch with an approach](https://github.com/vladmihalcea/hibernate-types/commit/e855c9c98a41d1d00ebecc5e1a9c29a82c0c3f49) for this issue as well but didn't have time to finish it.

lgoldstein:
It is difficult to provide a specific unit test, but I do have a use-case in mind - specifically, in a project I am involved with we are considering using JSONB to represent some compound entities we use. However, we have our own system-wide `ObjectMapper` that has some special settings, so I can define my own type mapper simply by extending yours and providing it with our JSON mapper in its default constructor the I described previously.

P.S. I have looked at the solution you suggest in the branch and I am fine with it as well - I do have some code remarks, but they can wait for now...

vladmihalcea:
It's not difficult at all. You can find many such test cases included in the project.

Since you want to provide a custom ObjectMapper, the test will demonstrate how you thought of declaring it because it can no longer be done via the @TypeDef annotation as it will require registering the Type during bootstrap.

lgoldstein:
Still - no problem -  as demonstrated in the code snippet I showed with _MyJsonBinaryType_

vladmihalcea:
I can't see any `MyJsonBinaryType` in your [Pull Request file changes](https://github.com/vladmihalcea/hibernate-types/pull/26/files).

vladmihalcea:
I had some time to implement this feature. Here's the [commit](https://github.com/vladmihalcea/hibernate-types/commit/fbc8dcc7fff9b4aa95aaa2350151bceea0138011). This will be supported from 2.1.0 onwards.

Thanks for your contribution.

lgoldstein:
Great - thanks

PR #28, opened 2018-03-19
Use Types.JAVA_OBJECT for AbstractJsonSqlTypeDescriptor, fixes #27

### LINKED ISSUE #27: Exception UNION could not convert type for inheritance hierarchies with JsonStringType
Given an class hierarchy with entities B and C that both extend base class A, when C has a property mapped with `JsonStringType`, and one queries all `A` using `SELECT a FROM A a`, the query fails with the following exception:

```
org.hibernate.exception.SQLGrammarException: could not extract ResultSet (3a7690e6-4840-4004-89d2-479af6a72963): javax.persistence.PersistenceException: org.hibernate.exception.SQLGrammarException: could not extract ResultSet
Caused by: org.hibernate.exception.SQLGrammarException: could not extract ResultSet
Caused by: org.postgresql.util.PSQLException: ERROR: UNION could not convert type json to uuid
  Position: 579
```

The error is caused by the fact that the produced SQL query has the following format: (I slightly simplified it to only include the relevant part)

```
SELECT b.*, null::uuid as properties FROM B b UNION ( select c.*, c.properties from C c)
```

The `::uuid` cast originates from [`PostgreSQL81Dialect#getSelectClauseNullString(sqlType)`](https://github.com/hibernate/hibernate-orm/blob/5.2.12/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java#L347-L361), which returns `uuid` for input `1111` (which is `Types.OTHER`). So it seems somehow, `OTHER` is already bound to `uuid`, and the work around performed by Hibernate does not work anymore (a better workaround would probably be to consider the columnDefinition of the column as well).

I'm uncertain whether this workaround is still required, but all new dialects (I am using the `PostgreSQL95Dialect`) still extend from the `PostgreSQL81Dialect` without overriding `getSelectClauseNullString` method. Furthermore, I am aware that this issue should probably be fixed in Hibernate rather than here. This method will in fact break for all queries using a null value of a type for which its JDBC type is mapped  to multiple types.

Nevertheless, I've had success with mapping my JSON properties to `Types.JAVA_OBJECT` rather than `Types.OTHER`. This type is not subject to the bug for me. `Types.JAVA_OBJECT` is also the JDBC type of choice as of the [`PostgreSQL92Dialect`](https://github.com/hibernate/hibernate-orm/blob/5.2.12/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL92Dialect.java#L23). 

Is there any particular reason to not switch the JDBC type of `JsonStringType` to `Types.JAVA_OBJECT`?

JWGmeligMeyling:
Suggested fix for #27 

Use `Types.JAVA_OBJECT` for `AbstractJsonSqlTypeDescriptor`, as it is the JDBC type that Hibernate uses itself in its dialects.

### DIFF
--- a/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/json/internal/AbstractJsonSqlTypeDescriptor.java
+++ b/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/json/internal/AbstractJsonSqlTypeDescriptor.java
@@ -18,7 +18,7 @@ public abstract class AbstractJsonSqlTypeDescriptor implements SqlTypeDescriptor
 
     @Override
     public int getSqlType() {
-        return Types.OTHER;
+        return Types.JAVA_OBJECT;
     }
 
     @Override
--- a/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/AbstractJsonSqlTypeDescriptor.java
+++ b/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/AbstractJsonSqlTypeDescriptor.java
@@ -18,7 +18,7 @@ public abstract class AbstractJsonSqlTypeDescriptor implements SqlTypeDescriptor
 
     @Override
     public int getSqlType() {
-        return Types.OTHER;
+        return Types.JAVA_OBJECT;
     }
 
     @Override
--- a/hibernate-types-5/src/main/java/com/vladmihalcea/hibernate/type/json/internal/AbstractJsonSqlTypeDescriptor.java
+++ b/hibernate-types-5/src/main/java/com/vladmihalcea/hibernate/type/json/internal/AbstractJsonSqlTypeDescriptor.java
@@ -18,7 +18,7 @@ public abstract class AbstractJsonSqlTypeDescriptor implements SqlTypeDescriptor
 
     @Override
     public int getSqlType() {
-        return Types.OTHER;
+        return Types.JAVA_OBJECT;
     }
 
     @Override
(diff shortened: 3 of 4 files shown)


vladmihalcea:
Thanks for the PR. I'll review it when I have time.

vladmihalcea:
I ran the tests and they fails:

```
Caused by: org.postgresql.util.PSQLException: Unsupported Types value: 2,000
	at org.postgresql.jdbc.PgPreparedStatement.setObject(PgPreparedStatement.java:682)
	at org.postgresql.jdbc.PgPreparedStatement.setObject(PgPreparedStatement.java:899)
```

It looks like the PostgreSQL driver does not support the `Types.JAVA_OBJECT`.

JWGmeligMeyling:
Peculiar, then I think it is dependent of the combinations of versions of Hibernate, dialect and Driver, and possibly even database that you're using. 

Under PostgreSQL 9.5, with driver 9.4-1202-jdbc41, Hibernate 5.2.12 and the Postgresql95Dialect,  I have better luck with the implementation suggested in https://stackoverflow.com/a/16049086/2104280 .

vladmihalcea:
I'm using 9.4-1202-jdbc41, PostgreSQL 9.6 and Postgresql95Dialect and tests are failing.

ehmPlankje:
The `Types.JAVA_OBJECT` type was only introduced in version 9.2 (see `org/hibernate/dialect/PostgreSQL92Dialect.java`). So if the driver uses anything older than that, it will fail. 

The issue mentioned in #27 also exists with `PostgreSQLIntervalType`, and there is no JDBC type for that one. I may have to use a different type than Interval, because Hibernate's table union is required in my domain model.

PR #31, opened 2018-05-11
Pass cause from ObjectMapper to wrapper clients

Sch3lp:
Otherwise a developer has to debug to discover that e.g. _No serializer found for class ...._ was the actual cause of _The given string value: ... cannot be transformed to Json object_.

Usually you want to know **why** it could not be transformed.


vladmihalcea:
Thanks for the PR. I will review it when I have some time.

vladmihalcea:
Applied. Thanks.

vladmihalcea:
The 2.2.1 version contains your changes and I've just released it. It will soon be available on Maven Central. Thanks and enjoy using it.

Sch3lp:
Thanks mate!

PR #40, opened 2018-08-08
Fix String fields with jsonb

### LINKED ISSUE #39: String fields on entity as JSONB don't work correctly
Hi,

we've recently discovered that String fields in entities don't work correctly with JSONB as a type. While data is persisted and can be read, jsonb operators like the ones from Postgres (e.g. `column->>'field'` ) won't work, which is a massive loss of functionality. The reason for this is that their not really persisted as valid JSONB. E.g. the String send to the database is the following:
```
"{\"country\":\"Romania\",\"city\":\"Cluj-Napoca\"}"
```
While it should be this:
```
{"country":"Romania","city":"Cluj-Napoca"}
```

The following patch should replicate the problem:
```
diff --git a/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonStringTypeTest.java b/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonStringTypeTest.java
index 30226a0..c40ee3d 100644
--- a/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonStringTypeTest.java
+++ b/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonStringTypeTest.java
@@ -1,5 +1,6 @@
 package com.vladmihalcea.hibernate.type.json;

+import com.vladmihalcea.hibernate.type.json.internal.JacksonUtil;
 import com.vladmihalcea.hibernate.type.model.BaseEntity;
 import com.vladmihalcea.hibernate.type.model.Location;
 import com.vladmihalcea.hibernate.type.model.Ticket;
@@ -15,6 +16,7 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;

 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;

 /**
  * @author Vlad Mihalcea
@@ -56,6 +58,7 @@ public class PostgreSQLJsonStringTypeTest extends AbstractPostgreSQLIntegrationT
             participant.setId(1L);
             participant.setTicket(ticket);
             participant.setEvent(event);
+            participant.setMetaData(JacksonUtil.toString(location));

             entityManager.persist(participant);

@@ -75,10 +78,17 @@ public class PostgreSQLJsonStringTypeTest extends AbstractPostgreSQLIntegrationT
                             "where p.ticket ->> 'price' > '10'")
                     .getResultList();

+            System.out.println(JacksonUtil.toString(location));
+            System.out.println(JacksonUtil.toString(JacksonUtil.toString(location)));
+            participant.setMetaData(JacksonUtil.toString(location));

             entityManager.persist(participant);

@@ -75,10 +80,18 @@ public class PostgreSQLJsonStringTypeTest extends AbstractPostgreSQLIntegrationT
                             "where p.ticket ->> 'price' > '10'")
                     .getResultList();

+            List<String> countries = entityManager.createNativeQuery(
+                    "select p.metadata->>'country' " +
+                            "from participant p ")
+                    .getResultList();
+
             event.getLocation().setCity("Constan<C8>");
             entityManager.flush();

             assertEquals(1, participants.size());
+            assertEquals(1, countries.size());
+            System.out.println(countries);
+            assertNotNull(countries.get(0));
         });
     }

@@ -110,6 +120,10 @@ public class PostgreSQLJsonStringTypeTest extends AbstractPostgreSQLIntegrationT
         @ManyToOne
         private Event event;

+        @Type(type = "jsonb")
+        @Column(columnDefinition = "jsonb")
+        private String metaData;
+
         public Ticket getTicket() {
             return ticket;
         }
@@ -125,6 +139,15 @@ public class PostgreSQLJsonStringTypeTest extends AbstractPostgreSQLIntegrationT
         public void setEvent(Event event) {
             this.event = event;
         }
+
+        public String getMetaData() {
+            return metaData;
+        }
+
+        public void setMetaData(String metaData) {
+            this.metaData = metaData;
+        }
+
     }
```

I believe the main reason for this is that an already valid JSON string goes through `JacksonUtil.toString()` again and thus ends up with backslashes in it.

Cheers,
Christoph

dreis2211:
Hi,

as discussed this is a first draft to fix #39.

Cheers,
Christoph

### DIFF
--- a/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonTypeDescriptor.java
+++ b/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonTypeDescriptor.java
@@ -84,6 +84,9 @@ public class JsonTypeDescriptor
 
     @Override
     public Object fromString(String string) {
+        if (String.class.isAssignableFrom((Class) type)) {
+            return string;
+        }
         return objectMapperWrapper.fromString(string, type);
     }
 
@@ -95,6 +98,9 @@ public class JsonTypeDescriptor
         }
         if (String.class.isAssignableFrom(type)) {
             return (X) toString(value);
+        }
+        if (Object.class.isAssignableFrom(type) && value instanceof String) {
+            return (X) objectMapperWrapper.toJsonNode((String) value);
         }
         if (Object.class.isAssignableFrom(type)) {
             return (X) objectMapperWrapper.toJsonNode(toString(value));
--- a/hibernate-types-43/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonBinaryTypeTest.java
+++ b/hibernate-types-43/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonBinaryTypeTest.java
@@ -1,5 +1,6 @@
 package com.vladmihalcea.hibernate.type.json;
 
+import com.vladmihalcea.hibernate.type.json.internal.JacksonUtil;
 import com.vladmihalcea.hibernate.type.model.BaseEntity;
 import com.vladmihalcea.hibernate.type.model.Location;
 import com.vladmihalcea.hibernate.type.model.Ticket;
@@ -13,6 +14,7 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 
 /**
  * @author Vlad Mihalcea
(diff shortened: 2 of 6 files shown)


vladmihalcea:
Thanks for the PR. I'll review it soon.

vladmihalcea:
@dreis2211 No need actually. This only happened if they encoded the JSON as String, which is a particular use case.

vladmihalcea:
Applied upstream, thanks.

dreis2211:
Thanks. Probably still worth mentioning it somewhere in the release notes (if there are any), don't you think?

vladmihalcea:
The idea is good, but I didn't have time to create a Release Notes page. If you want to it, feel free to send me a PR for it.

Issue #61, opened 2018-12-07
Pass array types in PostgreSQL functions as parameters in JPA query


StanislavGubanov:
Hello, Vlad.

How use array types in PostgreSQL database functions with array params in JPA

https://www.postgresql.org/docs/current/functions-array.html

function for example:
```sql
CREATE OR REPLACE FUNCTION fn_helper_array_contains(a integer[],b integer[])
RETURNS boolean AS
$$
BEGIN
  return a @> b;
END;
$$ LANGUAGE 'plpgsql';

-- select fn_helper_array_contains (array[1,2,3]::integer[],array[1,2]::integer[]);
```
###

vladmihalcea:
The ARRAY support in hibernate-types is for entity properties only. This is not supported with stored procedures as they don't take a Hibernate Type.

The only way to use ARRAY with stored procedures is via JDBC or using jOOQ.

vladmihalcea:
[This commit](https://github.com/vladmihalcea/hibernate-types/commit/e87179d65021024ec06c624535df3c67f39ba69e) shows how you can do it:

CriteriaBuilder cb = entityManager.getCriteriaBuilder();
CriteriaQuery<Event> cq = cb.createQuery(Event.class);
Root<Event> root = cq.from(Event.class);
cq.select(root);

```
ParameterExpression containValues = cb.parameter(int[].class, "arrayValues");
cq.where(cb.equal(cb.function("fn_helper_array_contains", Boolean.class, root.get("values"), containValues), Boolean.TRUE));
TypedQuery<Event> query = entityManager.createQuery(cq);
int[] arrayValues = {12, 16};
query.unwrap(Query.class).setParameter("arrayValues", arrayValues, IntArrayType.INSTANCE);
List<Event> events = query.getResultList();
assertArrayEquals(new int[]{12, 14, 16}, events.get(0).getValues());
```

Notice the `unwrap` to `org.hibernate.query.Query.class` and passing the `IntArrayType.INSTANCE` to bind the parameter:

    query.unwrap(Query.class).setParameter("arrayValues", arrayValues, IntArrayType.INSTANCE);

PR #62, opened 2018-12-07
add test for issue #61

StanislavGubanov:
example code for issue and this code get error, how can make work this
```org.postgresql.util.PSQLException: ERROR: function fn_helper_array_contains(integer[], bytea) does not exist```

### DIFF
--- a/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/array/ArrayTypeSearchTest.java
+++ b/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/array/ArrayTypeSearchTest.java
@@ -1,0 +1,112 @@
+package com.vladmihalcea.hibernate.type.array;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.fail;
+
+import com.vladmihalcea.hibernate.type.model.BaseEntity;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vladmihalcea** [comment]
Try unwrapping the `Query` to `org.hibernate.query.Query` and call the `setParameter` method that takes a Hibernate `Type` and pass the IntArrayType.

**Comment by vladmihalcea** [comment]
Applied upstream and fix the test as indicated in the comment above:

```
CriteriaBuilder cb = entityManager.getCriteriaBuilder();
CriteriaQuery<Event> cq = cb.createQuery(Event.class);
Root<Event> root = cq.from(Event.class);
cq.select(root);

ParameterExpression containValues = cb.parameter(int[].class, "arrayValues");
cq.where(cb.equal(cb.function("fn_helper_array_contains", Boolean.class, root.get("values"), containValues), Boolean.TRUE));
TypedQuery<Event> query = entityManager.createQuery(cq);
int[] arrayValues = {12, 16};
query.unwrap(Query.class).setParameter("arrayValues", arrayValues, IntArrayType.INSTANCE);
List<Event> events = query.getResultList();
assertArrayEquals(new int[]{12, 14, 16}, events.get(0).getValues());

It works like a charm!
```

PR #63, opened 2018-12-15
Add support for PostgeSQL range type.

nstdio:


vladmihalcea:

vladmihalcea:
@nstdio I integrated it. Thanks.

nstdio:
@vladmihalcea Thanks. Would it be appropriate to make `Range` API more rich ?

For example some how expose/adapt [this functions](https://www.postgresql.org/docs/current/functions-range.html) at Java level.

vladmihalcea:
All those operators can be accessed via native SQL. The goal of this framework is just to allow Hibernate to persist database-specific column types.

AbdulR3hman:
@vladmihalcea any plans on releasing this soon?

vladmihalcea:
@AbdulR3hman I'll integrate [this PR first](https://github.com/vladmihalcea/hibernate-types/pull/64),and then release a new version in the following two weeks.

PR #68, opened 2018-12-27
Fix ClassCastException for PostgreSQL enum array

nkonev:
Hello.
I got ClassCastException during instantiate array.

There is stacktrace.
```
Caused by: java.lang.ClassCastException: class org.postgresql.util.PGobject cannot be cast to class java.lang.String (org.postgresql.util.PGobject is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
	at com.vladmihalcea.hibernate.type.array.internal.ArrayUtil.unwrapArray(ArrayUtil.java:161) ~[hibernate-types-52-2.3.5.jar:na]
	at com.vladmihalcea.hibernate.type.array.internal.AbstractArrayTypeDescriptor.wrap(AbstractArrayTypeDescriptor.java:70) ~[hibernate-types-52-2.3.5.jar:na]
	at com.vladmihalcea.hibernate.type.array.internal.ArraySqlTypeDescriptor$2.doExtract(ArraySqlTypeDescriptor.java:55) ~[hibernate-types-52-2.3.5.jar:na]
	at org.hibernate.type.descriptor.sql.BasicExtractor.extract(BasicExtractor.java:47) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:261) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:257) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:247) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
	at org.hibernate.type.AbstractStandardBasicType.hydrate(AbstractStandardBasicType.java:333) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
	at org.hibernate.persister.entity.AbstractEntityPersister.hydrate(AbstractEntityPersister.java:3007) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
	at org.hibernate.loader.plan.exec.process.internal.EntityReferenceInitializerImpl.loadFromResultSet(EntityReferenceInitializerImpl.java:305) ~[hibernate-core-5.3.7.Final.jar:5.3.7.Final]
[...]

PR #71, opened 2019-01-03
Fix & add tests

nkonev:
Hello, as you asked [here](https://github.com/vladmihalcea/hibernate-types/pull/68#pullrequestreview-188212255) I added positive test for PostgreSQL's enum array.
Also I added docker-compose files for quickly setup test environment with MySQL & PosgreSQL.
Also I fixed test failures due singleton caching old properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,3 +99,20 @@ Depending on the Hibernate version you are using, you need to following dependen
 * SLF4J
 * Jackson Databind
 
+## How to quickly start test environment
+```bash
+cd docker
+docker-compose up -d
+```
+
+## How to quickly access to database logs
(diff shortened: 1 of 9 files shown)


vladmihalcea:
@nkonev Thanks for the PR. You need to split it into 2 PRs:

- the first PR has to be the one which adds the test case that I requested
- the second PR should be the one with Docker Compose

nkonev:
@vladmihalcea ok, I will split It to two PR:

first with tests (fixes existing and add new),

second with Docker Compose.

vladmihalcea:
It's correct. Looking forward to it.

PR #75, opened 2019-01-03
change constructor type args from Class to Type

andyflury:
The JsonStringType constructor args Class are only used when calling the constructor of JsonTypeDescriptor which expects a Type argument. There is no need for the JsonStringType constructor arg to be more restrictive (Class) then the JsonTypeDescriptor arg (Type). 

This causes an issue when grammatically registering the JsonStringType and injecting a java.lang.reflect.ParameterizedType (e.g. HashMap<String,Object>)

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/json/JsonStringType.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/json/JsonStringType.java
@@ -5,9 +5,11 @@ import com.vladmihalcea.hibernate.type.json.internal.JsonStringSqlTypeDescriptor
 import com.vladmihalcea.hibernate.type.json.internal.JsonTypeDescriptor;
 import com.vladmihalcea.hibernate.type.util.Configuration;
 import com.vladmihalcea.hibernate.type.util.ObjectMapperWrapper;
+
 import org.hibernate.type.AbstractSingleColumnStandardBasicType;
 import org.hibernate.usertype.DynamicParameterizedType;
 
+import java.lang.reflect.Type;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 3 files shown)


vladmihalcea:
In order to integrate this change, I need a test case that proves why we need this modification. Therefore, you need to amend your PR to include the same test case you described in the issue that led to this change request.

andyflury:
I've add MySQLGenericConfigJsonTypeTest. It programatically registers the `JsonStringType` using a ParameterizedType (HashMap<String,Object>). It adds a new Entity `PropertyHolder` that contains a HashMap property (`props`) that is persisted into a json column. The `props` property can hold objects as well as POJO.

Without the Constructor changes this test case will fail.

Feel free to make any changes you find necessary.

vladmihalcea:

andyflury:
Thanks! I added a ParametrizedJsonStringType subclass so that it can be used alongside the default JsonStringType

andyflury:
@vladmihalcea I believe I figured out away to get around the constructor arg problem by having my ParametrizedJsonStringType extend AbstractSingleColumnStandardBasicType instead of JsonStringType:

```
public class ParametrizedJsonStringType extends AbstractSingleColumnStandardBasicType<Object> implements DynamicParameterizedType {

    public ParametrizedJsonStringType(final String name, final Type rawType, final Type[] actualTypeArguments) {
        super(
                JsonStringSqlTypeDescriptor.INSTANCE,
                new JsonTypeDescriptor(new ObjectMapperWrapper(createObjectMapper()), createParameterizedType(rawType, actualTypeArguments, null))
        );
    }
```

With this my solution works without any modifications of your library.

Nevertheless I think It would be good to change the constructor args of JsonStringType to Type instead of Class. Up to you!

Thanks for your help along the way!

vladmihalcea:
Applied upstream, thanks

PR #81, opened 2019-02-20
Created tsvector user type for hibernate 5.2

bitluke:
Please review the changes for tsvector usertype.


vladmihalcea:
I think it would be more valuable if, instead of a `TSVector` class, a `String` could be used on the Java side. For this, you need to take a look at the JSON types and use the SQL/Java descriptors. The `UserType` is much more limited and does not provide this functionality.

PR #82, opened 2019-02-20
Add a LongArrayType to support long[] array mappings

pmclaren:
With this PR I added support for long arrays based on the implementation of int arrays I also added tests. Please feel free to edit whatever is necessary.


vladmihalcea:
Applied upstream, thanks

PR #86, opened 2019-03-18
Issue with ZoneDateTime when valid milliseconds is less than 4 or less decimal places

arunmohandas196:
Hi there,
Great project!
Wanted to highlight a problem I have faced while using this for tstzrange.

For tstzrange, if the ZoneDateTime creation has valid milliseconds less than 4 or less decimal places, retrieval from db fails for the persisted records.

Sample dateTime values which gives error:
"2018-05-03T10:15:30.127100+12:00"
""2018-05-03T10:15:30.127000+12:00"

Even though  ZONE_DATE_TIME has format "yyyy-MM-dd HH:mm:ss[.SSSSSS]X"), while persisting to postgres it ignores trailing zeros.
Retrieval of record fails during formating this.
Eg:
"java.time.format.DateTimeParseException: Text '2018-05-03 00:15:30.1271+02' could not be parsed at index 10"

Sample test case added in the pull request. Please review.


vladmihalcea:
Thanks for the test case. I'll take a look at it when I have some time.

calvequi:
Hi @vladmihalcea! I think pull request  #90 solved this issue.

vladmihalcea:
Applied upstream, thanks

PR #87, opened 2019-03-21
Implemented Postgres TSVector

bitluke:
I have just implemented TsVector using the descriptor type . 

Basically, I implemented it in two ways : one with dynamic parameters (which implements DynamicParameterizedType) and the other non dynamic parameters (which does not implement DynamicParameterizedType)

For dynamic parameters they can be found in the package tsvector.dynamicparameters 
For non dynamic parameters they can be found in the package tsvector.nondynamicparameters 

I had to include pg driver at compile time because I needed to be able to see PgObejct since Postgres abstract the concept of non supported JDBC types like tsvector.

Could you please review and tell which choice you prefer then I can now finally replicate to 5, 4, 4.2 module ?


vladmihalcea:
Cool, thanks. Looking forward to reviewing it.

vladmihalcea:
There are multiple issues related to your commit:

1. First of all, there is a merge commit. You need to create a new branch and cherry-pick all the commits so that I can rebase and squash them.
2. You don't need the `TSVector`. I think it's better to just use a `String` instead on the entity side.
3. Don't use the `PGobject` at compile time. Check out the `PostgreSQLHStoreType` to see how you can address this without needing to link the class. Maybe some users need some type just for MySQL so it's not good to have them require the PostgreSQL dependency.
4. Using the `PGobject` in the JavaDescriptor is not a good idea either. There you only want to have standard Java types that are not linked to any SQL-specific JDBC Driver. Maybe you send the entity to the web layer.
5. And `PGTSVector` is also not needed.

So, remove everything in the `nondynamicparameters` package because it's not needed. So, you need to get the Object as a String from the driver and just use it like that in the Java type descriptor.

rieckpil:
@bitluke are you still working on this or can I take over? I currently need the data type for a project and would implement it

vladmihalcea:
@rieckpil Cool. I think that on the Java side, we should be using `String` only for the moment.

rieckpil:
Using a String will result in the following: `"'Lorem' 'amet' 'dolor' 'ipsum' 'sit'"`. Would you leave it like this or replace the single quotes?

vladmihalcea:
Then maybe mapping the TSVector to a String array would make more sense. Frankly, I haven't used it so I don't know the best way to map it.

rieckpil:
I think working with a `String[]` within Java would be a little bit cumbersome (e.g. when inserting new rows, the code has to split the text into the correct form. Do you think `.replace("'", "")` is good enough for a start?

bitluke:
Sorry about my late response. 

@vladmihalcea. Postgresql is only open source db that has tsvector functionality implemented [https://en.wikipedia.org/wiki/Full-text_search]. Full text search implementation in postgresql actually makes use of a some functions like tsquery, tsvector etc. hence, I think it would be better to have a class that extend the Pgobject that is specially dedicated to Postgres Objects that is not supported by JDBC or even the common open source Db existing . What do you think  ? 

 @rieckpil if you want I to take over the implementation I can brief you on the implementation i have done.

rieckpil:
@bitluke I'm currently struggeling to represent the `tsvector` correctly within Java, using the value from postgres as plain Java `String` will result in the following: 
`
"'age':50 'anoth':41 'bed':54 'believ':2 'comfort':7 'convict':43 'cousin':11 'day':28,55 'delight':10 'desir':29 'discret':48 'enabl':37 'end':18 'evid':8 'expens':31 'expos':39 'feel':13,24 'find':25 'full':52 'gate':53 'genius':15 'inform':30 'invit':38 'learn':27 'look':17 'lose':56 'man':26 'materi':32 'minut':14 'mr':46 'piqu':19 'reason':42,49 'request':4 'return':33 'rose':23 'six':34 'solicitud':44 'spring':20"
`

I thought of reimplementing PostgreSQL's `tsvector_to_array(tsvector)` in Java and use this result, but once the text is stored as `tsvector` I think reproducing the original english text is not possible. 

What do you think?

bitluke:
The real issue is that you the tsvector type is shown as a string but in the true sense it is not i.e a string in java always has a quote like "this is my age" meanwhile tsvector is like a compressed string and it is represented as >> 'age': 4 'my':2 << you can see i did not use the quotations. I got to learn about this the hard way :).  Also "this and  is" are omitted.  

Back to your issue, keep in mind that you can not build the text back but it can be updated. tsvector is like a representation of possible strings you would like to search for. That is if I have a row that contains so three unique columns and I want to be able to search for those columns at lighting speed . tsvector representation makes it easier to do that instead of running a query that looks like 

`select * from where col1 like '%abc%' and col2 like '%cde%' and col3 like '%gfh%'; `

So in essence, you can store and update the value via a Postgres trigger as implemented in sample  [pagila](https://www.postgresql.org/ftp/projects/pgFoundry/dbsamples/pagila/pagila/) database. (just download and see how the trigger is used to update the column). I see no use of getting the reading the value.

Summary Tsvector is just to facilitate full text search in Postgressql. Which means you need not overkill your app with luecene or solr in order to implement search features.

Please note i am not an authority on the subject but these are my findings :). I hope it helps ? ;)

bitluke:
Just to add to your last comment. Looking at 
[the doc ](https://www.postgresql.org/docs/9.6/functions-textsearch.html) there is says the following.

tsvector_to_array(tsvector) -> convert tsvector to array of lexemes 

Keep in mind lexemes are just mapped string and not the real strings :)

rieckpil:
Thanks for your response. I had a wrong perception of what the `tsvector` type really is. So using this for a demo use case would mean to duplicate the information e.g. a description of a `Product` in two columns: one for UI representation (the full product description in readable english) and one column just for optimized searches with a `tsvector` column and using the trigger for keeping both in sync. Or is there a better way to use it?

bitluke:
@rieckpil  Yes, I think you now understand the idea.

Just some correction, as regards your description column, I don't think you are duplicating entries . For example if description is "a fat  cat sat on a mat - it ate a fat rats"  and you call `to_tsvector(description) ` you would get  `'ate':9 'cat':3 'fat':2,11 'mat':7 'rat':12 'sat':4`

As you can see this is not a duplication. it is just like a compressed description without `"it and a"`. The numbers represent the location of the word in the string. I personally see no use of TSvector on the UI except for internal use with [functions stated in the docs](https://www.postgresql.org/docs/9.6/functions-textsearch.html)

The triggers that can be called include: `tsvector_update_trigger()` and `tsvector_update_trigger_column()`

Still interested in the implementation ?

rieckpil:
I would then keep the implementation as simple as the following: https://github.com/rieckpil/hibernate-types/commit/b14a28710e76fafd4a9ed84cdec7fd5ef5863d5d

rieckpil:
So you wouldn`t use `tsvector` in your database and always use `varchar()` or `text` and use `to_tsvector(column)` on every FTS call to make use of the FTS capability? 

What I meant with duplication was having two columns, one with the `varchar()` and one with `tsvector` so the lexemes can be created during the insert and not on-demand for every query (don't know how resource intensive the creation adhoc is)

bitluke:
I am going to use tsvector all I am just saying there is no need to display the content to a user except if the user really needs the metadata. 

Two ways of updating the data that I can think of now is the application of trigger and possible calling the a native query to get the value and then updating the specific attribute. 

Yes to this `rieckpil@b14a287` but keep in mind the suggestion @vladmihalcea  mentioned.


@vladmihalcea do you have any objections based on the discussion in this thread ?

vladmihalcea:
@bitluke I don't have experience with TSVector. Yet, I don't see why would want to expose the entity field as `PgObject`. That does not sound right. Why wouldn't it work with a `String` instead and do the translation in the Java wrapper?

rieckpil:
I'll create the pull request this weekend. @vladmihalcea should I add the type to all `hibernate-type` versions?

vladmihalcea:
@rieckpil Yes, please. Unless it's a Java 1.8 based Types, we should have all Types added to all Hibernate versions.

vladmihalcea:
Superseded by [this PR](https://github.com/vladmihalcea/hibernate-types/pull/91).

PR #91, opened 2019-04-07
Add PostgreSQL TSVector type 

rieckpil:
I've added PostgreSQL `TSVector` data type with a Java `String` representation. Please have a look at the test (https://github.com/vladmihalcea/hibernate-types/pull/91/commits/67bef47a7b2411d100492df1f62e5a1a4189dabe)  for the new type. I would add the missing tests to the other `hibernate-typesXY` folder if the test is okay.

### DIFF
--- a/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/basic/PostgreSQLTSVectorType.java
+++ b/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/basic/PostgreSQLTSVectorType.java
@@ -2,7 +2,7 @@ package com.vladmihalcea.hibernate.type.basic;
 
 import com.vladmihalcea.hibernate.type.ImmutableType;
 import com.vladmihalcea.hibernate.type.util.ReflectionUtils;
-import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.engine.spi.SessionImplementor;
 
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
@@ -15,7 +15,7 @@ import java.sql.Types;
  * @author Vlad Mihalcea
  * @author Philip Riecks
  */
-public class PostgreSQLTSVectorType  extends ImmutableType<String> {
+public class PostgreSQLTSVectorType extends ImmutableType<String> {
 
     public PostgreSQLTSVectorType() {
         super(String.class);
@@ -27,13 +27,13 @@ public class PostgreSQLTSVectorType  extends ImmutableType<String> {
     }
 
     @Override
-    protected String get(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner)
+    protected String get(ResultSet rs, String[] names, SessionImplementor session, Object owner)
             throws SQLException {
         return rs.getString(names[0]);
     }
 
     @Override
-    protected void set(PreparedStatement st, String value, int index, SharedSessionContractImplementor session)
+    protected void set(PreparedStatement st, String value, int index, SessionImplementor session)
             throws SQLException {
         if (value == null) {
             st.setNull(index, Types.OTHER);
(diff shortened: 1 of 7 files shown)


vladmihalcea:
The PR looks good, but I would not use the `UserType` hierarchy for this type. Having it as a `String` on the Java side is fine, but now this is your only option.

On the other hand, if you use the Java and SQL Type Descriptors as the JSON Types do, you have the flexibility of varying the Java-side entity property.

For instance, considering we have this `String` based TSVector:

`'access':9,35 'batch':16 'book':2 'common':32 'concurr':21 'connect':13 'control':22 'data':8,34 'fetch':18 'framework':36 'inner':27 'java':7,33 'journey':5 'manag':14 'mechan':23 'perform':10 'size':19 'tune':11 'unravel':25 'updat':17 'work':28`

We could use a `Map` on tha Java-side for it, and the Java Descriptor can allow us to have both `Map` and `String` while using the same SQL Type descriptor.

Therefore, the PR needs some changes in order to be integrated.

rieckpil:
Okay so the TSVector type should `extends AbstractSingleColumnStandardBasicType<Object> implements DynamicParameterizedType `? With the String based TSVector you posted, only a `Map<String, List<Integer>>` would make sense, wouldnt't it? 

Which JSON type should I have a look for a reference? `JsonBinaryType`, `JsonStringType`, `JsonNodeBinaryType` or `JsonNodeStringType`?

vladmihalcea:
@rieckpil Yes, it should extend `AbstractSingleColumnStandardBasicType`. A `Map<String, List<Integer>>` amkes sense indeed.

The `JsonBinaryType` is a good place to take a look. The Java descriptor is the place where you adapt the Java-side Object type to the one you pass to the SQL Descriptor. That's why it's more flexible because you can have multiple Object types on the Java side.

rieckpil:
@vladmihalcea I am currently a little bit confused as I don't know exactly what you mean. I tried to add a basic setup with my last commit but I am not sure about it. 

Do you have a link to the Hibernate docs which could help me improve my knowledge about this?

vladmihalcea:
There is no Hibernate documentation for this. However, if you follow the JsonBinaryType, you will see there are tests that show it works for both custom DTOs or String on the entity property side.

If you cannot get it working, I'll try to do it sometime in the future. However, for the moment, my Hypersistence Optimizer is a Top priority for me, so I will investigate it in one or two months.

JWGmeligMeyling:
> The PR looks good, but I would not use the UserType hierarchy for this type. Having it as a String on the Java side is fine, but now this is your only option.

I think implementing it as an `ImmutableType` now, would not really restrict the later change to an `AbstractHibernateType`? Any existing mapped string would still map appropriately even if the implementation changes right?

> We could use a `Map` on tha Java-side for it, and the Java Descriptor can allow us to have both Map and String while using the same SQL Type descriptor.

I'm not sure whether a `Map` will suffice. First, I think the `tsvectors` are ordered, so that would probably need a `LinkedHashMap` as a basis. Secondly, the positions may also have an included weight. At that point we're better off building a full TSVector bean type.

My use case is actually not even to be able to read or write to the tsvector field, but just to be able to map it in my entities and reference it from criteria queries in order to do text search. I don't actually care which Java representation it has, because I only reference it from queries. I think commit https://github.com/vladmihalcea/hibernate-types/pull/91/commits/e572e201ab6b25a7ed3facccddce1f3820d1dd62, that implements the `TSVectorType` as an `ImmutableType`, is good enough for this purpose, and I don't really see how another implementation based on a `AbstractSingleColumnStandardBasicType` would improve things for now. Is there any chance that that implementation could be merged? Possibly after the introduction of a couple of tests?

vladmihalcea:
Ok. I'll review it one more time and probably include it in the next release.

vladmihalcea:
@rieckpil @JWGmeligMeyling This is [the Type using SQL and Java descriptors](https://github.com/vladmihalcea/hibernate-types/pull/116). However, as discussed in the PR, I think this is better achieved via a separate column populated by PostgreSQL.

PR #99, opened 2019-05-26
Replace reflection usage in PostgreSQL Types with Driver-specific API calls

nstdio:


vladmihalcea:
Reflection usage is intentional. Without it, a mandatory dependency on PostgreSQL Driver will be needed, and that's not desirable since the framework can be used for MySQL only.

nstdio:
I don't understand how it's mandatory? We add postgres driver as a provided dependency...

vladmihalcea:
Ok, I'll investigate it then.

vladmihalcea:
Applied. Thanks.

Issue #100, opened 2019-05-29
In case the JSON attribute is a java.util.Set, an additional SQL UPDATE is executed when persisting the entity


Cepr0:
If the JSON attribute in the entity is `Set`, like the following for example:
```java
@Entity
@Table(name = "users")
@TypeDef(name = "jsonb", typeClass = JsonBinaryType.class)
public class User {
    @Type("jsonb")
    @Column(columnDefinition = "jsonb")
    private Set<String> phones;
    // ...
}
```
then the extra update is performed when we create a new entity:
```sql
insert into users (version, phones, id) values (?, ?, ?)
update users set version=?, phones=? where id=? and version=?
```
This issue occurs regardless of the type parameter of the `Set` - either simple type, enum or POJO. It occurs with `EnumSet` as well. 

The issue can be fixed by adding the following check to the begging of `areEqual` method in `JsonTypeDescriptor` class:
```java
public class JsonTypeDescriptor
        extends AbstractTypeDescriptor<Object> implements DynamicParameterizedType {
    //...
    @Override
    public boolean areEqual(Object one, Object another) {
        if (one instanceof Collection && another instanceof Collection) {
            return Objects.equals(one, another);
        }
        //...
    }
}
```
If such a fix is correct I would be able to create a corresponding PR.

vladmihalcea:
Please send a Pull Request with the change so that I can debug it too. Thanks.

Cepr0:
Have done: [#101](https://github.com/vladmihalcea/hibernate-types/pull/101)

vladmihalcea:
Thanks. I'll review it when I have some time.

vladmihalcea:
Fixed.

PR #101, opened 2019-05-29
In case the JSON attribute is a java.util.Set, an additional SQL UPDATE is executed when persisting the entity #100

Cepr0:


vladmihalcea:
Applied. Thanks.

Cepr0:
Glad to help!

PR #106, opened 2019-06-24
Add handling for null PostgreSQL Range column values

neetkee:
It appears that NPE throws while getting null ranges from a table. PR tries to handle this situation.


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

PR #110, opened 2019-07-10
Type implementation for PostgreSQL interval, fixes #53

### LINKED ISSUE #53: Add support for PostgreSQL Interval column type
Hi Vlad,

I have a custom type for PGIntervals that I map to a Java duration. This is for example useful to map result sets of native queries that subtract one timestamp for another. Let me know if there's interest in adding it to the custom type library:

```java
public class PGIntervalType implements UserType {

    public static final PGIntervalType INSTANCE = new PGIntervalType();

    public Class<?> returnedClass() {
        return Duration.class;
    }


    public int[] sqlTypes() {
        return new int[] {Types.OTHER};
    }


    @Override
    public Object nullSafeGet(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner) throws HibernateException {
        try {
            final PGInterval pgi = (PGInterval) rs.getObject(names[0]);
            final int days = pgi.getDays();
            final int hours = pgi.getHours();
            final int mins = pgi.getMinutes();
            final double secs = pgi.getSeconds();

            return Duration.ofDays(days)
                .plus(hours, ChronoUnit.HOURS)
                .plus(mins, ChronoUnit.MINUTES)
                .plus(Math.round(secs), ChronoUnit.SECONDS);

        }
        catch (Exception e) {
            return null;
        }
    }


    @Override
    public void nullSafeSet(PreparedStatement st, Object value, int index, SharedSessionContractImplementor session) throws HibernateException, SQLException {
        if (value == null) {
            st.setNull(index, Types.OTHER);
        }
        else {
            final Duration period = ((Duration)value);
            final int days = (int) period.toDays();
            final int hours = (int) (period.toHours() - days * 24);
            final int mins = (int) (period.toMinutes() - period.toHours() * 60);
            final int secs = (int) (period.getSeconds() - mins * 60);

            final PGInterval pgi = new PGInterval(0, 0, days, hours, mins, secs);
            st.setObject(index, pgi);
        }
    }


    public boolean equals(Object x, Object y)
        throws HibernateException {

        return x == y;
    }


    public int hashCode(Object x)
        throws HibernateException {
        return x.hashCode();
    }


    public Object deepCopy(Object value)
        throws HibernateException {
        return value;
    }


    public boolean isMutable() {
        return false;
    }


    public Serializable disassemble(Object value)
        throws HibernateException {
        return (Serializable) value;
    }


    public Object assemble(Serializable cached, Object owner)
        throws HibernateException {
        return cached;
    }


    public Object replace(Object original, Object target, Object owner)
        throws HibernateException {
        return original;
    }

}
```

JWGmeligMeyling:
Implements a custom type for the PostgreSQL Interval type, and maps it to a Java 8 Duration (see also: https://github.com/vladmihalcea/hibernate-types/issues/53).

Notes:
- Only implemented in `hibernate-types-52` due to the dependency on Java 8
- Nanos are not considered because the value would lose precision in the long to float conversion anyway

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/interval/PostgreSQLIntervalType.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/interval/PostgreSQLIntervalType.java
@@ -1,0 +1,65 @@
+package com.vladmihalcea.hibernate.type.interval;
+
+import com.vladmihalcea.hibernate.type.ImmutableType;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.postgresql.util.PGInterval;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Types;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Maps an {@link Duration} object type to a PostgreSQL Interval column type.
+ * <p>
+ *
+ * @author Jan-Willem Gmelig Meyling
+ */
+public class PostgreSQLIntervalType extends ImmutableType<Duration> {
+
+    public static final PostgreSQLIntervalType INSTANCE = new PostgreSQLIntervalType();
+
+    public PostgreSQLIntervalType() {
+        super(Duration.class);
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


vladmihalcea:
Thanks. I'll take a look when I have some time.

vladmihalcea:
Applied upstream, thanks

PR #116, opened 2019-08-15
Add support for PostgreSQL tx_vector column type

vladmihalcea:
This PR supersedes the [original work](https://github.com/vladmihalcea/hibernate-types/pull/91) done by @rieckpil.

However, having a Hibernate Type is not as powerful as just using a TRIGGER that populates the ts_vector column by calling `to_tsvector`. The current implementation only casts the provided String to a `ts_vector` type which lacks the integer positions of the associated lexemes.

### DIFF
--- a/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/basic/PostgreSQLTSVectorType.java
+++ b/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/basic/PostgreSQLTSVectorType.java
@@ -1,0 +1,47 @@
+package com.vladmihalcea.hibernate.type.basic;
+
+import com.vladmihalcea.hibernate.type.ImmutableType;
+import com.vladmihalcea.hibernate.type.util.ReflectionUtils;
+import org.hibernate.engine.spi.SessionImplementor;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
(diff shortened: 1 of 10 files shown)


JWGmeligMeyling:
@vladmihalcea I see you recently updated this branch, what are the changes you still want to make for this branch?

vladmihalcea:
I tested it to see how it works, and realized that having this Type won't be very effective since the `ts_vector` is better off populated via the `to_tsvector` SQL function call.

JWGmeligMeyling:
I don't think the type will be used for manipulation often. In my case though I am mostly interested in just being able to map it and type it for custom SQL functions. Often you do not want to convert a single column into a tsvector, but is your tsvector comprised from data from several columns or possibly even separate tables. In my case I have a materialized view that contains the tsvector that I map in Hibernate to an immutable view through a OneToOne relation. This type, as limited as it is, would allow me to map the tsvector column as attribute, allowing me to pass it as argument in JPQL functions.

If we're not comfortable with a more obscure type implementation that just puts the value from any type of `PGObject` in a `String` value: i.e. `PGObjectAsStringType`.

vladmihalcea:
If you think it's worth having it, then I'm going to integrate it as is.

JWGmeligMeyling:
Let me test your type locally. I'll add a review after I have confirmed it works the way I intend it!

vladmihalcea:
That would be great, thanks.

PR #122, opened 2019-08-28
Adds support of NonNullStringType

akinchev:
Usually string-based fields have two possible values for “no data”: NULL, and the empty (blank) string.
In most cases, it’s redundant and can lead to mistakes.
This PR adds support for NonNullStringType, which helps to avoid two possible values for “no data” by replacing NULLs with blank lines.

### DIFF
--- a/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/basic/NonNullStringType.java
+++ b/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/basic/NonNullStringType.java
@@ -1,0 +1,48 @@
+package com.vladmihalcea.hibernate.type.basic;
+
+import com.vladmihalcea.hibernate.type.ImmutableType;
+import org.hibernate.engine.spi.SessionImplementor;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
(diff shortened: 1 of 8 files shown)


vladmihalcea:
Why not use the `@NotNull` annotation instead? I don't think that input validation should be done by the Hibernate Type. Bean Validation already covers this responsibility.

akinchev:
Sure. It is possible to resolve the problem of two "no data" values of string based fields via validation, but:

1. Your application can be not the only source of data in your database (I know that it's not a good practice, but unfortunately I've seen quite a lot of projects where some data synchronisation between services was made via database).
2. Solving it through validation may lead you to multiple validation checks of DTOs cause there maybe quite a lot of API versions.

The philosophy of this solution is, that client sends "no data" string value to the application not caring about the format of this "no data" value (just like date and time values which can be accepted in any reasonable format).

vladmihalcea:
Validation is needed on every layer boundary. So, you have it in the UI, Web Controllers, DAO layer and the DB. This will ensure that the failure happens as soon as possible and will allow reusing base layers on newer UI models like Mobile clients.

However, I don't see the value in having the Hibernate Types do validation. That shpuld be done way prior to flushing the entities. Hence, I'm closing this issue.

PR #125, opened 2019-09-11
Ability to use PostgreSQLEnumType and EnumArrayType with TypedParameterValue

### LINKED ISSUE #124: Initialize PostgreSQLEnumType programmatically
I am trying to create a `TypedParameterValue` (from your blogpost https://vladmihalcea.com/bind-custom-hibernate-parameter-type-jpa-query/ ) for a `PostgreSQLEnumType`. In order to do so I use:

```java
PostgreSQLEnumType enumType = new PostgreSQLEnumType();
Properties properties = new Properties();
properties.put("enumClass", EnumClass.class.getName());
properties.put("useNamed", true);
enumType.setParameterValues(properties);
Enum enumValue = ...
query.setParameter("param", new TypedParameterValue(enumType, enumValue));
```

Obivously, the properties have to be set, because `EnumType`, the base type for `PostgreSQLEnumType` is a `DynamicParameterizedType`, and requires the enum classname to be set through properties for it not to fail with an NPE.

However, invoking `setParameterValues` requires `typeConfiguration`, which is set through `setTypeConfiguration`, as the type is `TypeConfigurationAware`. Because our type is instantiated programmatically, this `typeConfiguration` is `null`.

A couple of questions:

1. Can we add an additional constructor to `PostgreSQLEnumType` of the format `PostgreSQLEnumType(Class<Enum> clasz)`? However, the most tricky part of this:
2. How to get hold of a `TypeConfiguration` object?
3. Or: can the `PostgreSQLEnumType` be rewritten so that the `TypeConfiguration` is not required anymore? (We would have to replace the manner of construction of the `EnumJavaTypeDescriptor`, but in order to do so probably end up override all of the `EnumType` implementation - as its private).

I have also thought of registering the type globally and looking it up somewhere from Hibernate, but so far I haven't been able to get this to work.

I think we can also conclude that the fact that `PostgreSQLEnumType` is a `TypeConfigurationAware DynamicParameterizedType` renders the presence of the singleton `PostgreSQLEnumType.INSTANCE` useless, as it cannot be used in that fashion.

JWGmeligMeyling:
This allows the instantation of `PostgreSQLEnumType`  and `EnumArrayType` with the parameters normally set through `DynamicParameterizedType` set through the constructor. This allows for the `PostgreSQLEnumType`  and `EnumArrayType` to be used with [`org.hibernate.Query#setParameter(int position, Object val, Type type)`](https://docs.jboss.org/hibernate/orm/5.4/javadocs/org/hibernate/Query.html#setParameter-int-java.lang.Object-org.hibernate.type.Type-) or as [`TypedParameterValue`](https://docs.jboss.org/hibernate/orm/5.4/javadocs/org/hibernate/jpa/TypedParameterValue.html). As both methods are introduced in Hibernate 5.x, the changes are only ported to the modules targeting Hibernate 5.x. This applies both my fix and the suggestion to stub the `TypeConfiguration`, as both are required to make the test work. I also had to stub `DynamicParameterizedType.ParameterType` for the `EnumArrayType` - this could also be done differently, but this requires minimal changes to the types themselves.

Fixes #124

### DIFF
--- a/hibernate-types-5/src/main/java/com/vladmihalcea/hibernate/type/array/EnumArrayType.java
+++ b/hibernate-types-5/src/main/java/com/vladmihalcea/hibernate/type/array/EnumArrayType.java
@@ -4,6 +4,7 @@ import com.vladmihalcea.hibernate.type.AbstractHibernateType;
 import com.vladmihalcea.hibernate.type.array.internal.ArraySqlTypeDescriptor;
 import com.vladmihalcea.hibernate.type.array.internal.EnumArrayTypeDescriptor;
 import com.vladmihalcea.hibernate.type.util.Configuration;
+import com.vladmihalcea.hibernate.type.util.ParameterizedParameterType;
 import org.hibernate.annotations.Type;
 import org.hibernate.usertype.DynamicParameterizedType;
 
@@ -41,6 +42,14 @@ public class EnumArrayType
             new EnumArrayTypeDescriptor(),
             configuration
         );
+    }
+
+    public EnumArrayType(Class<? extends Enum> enumClass, String sqlArrayType) {
+        this();
+        Properties parameters = new Properties();
+        parameters.setProperty(SQL_ARRAY_TYPE, sqlArrayType);
+        parameters.put(DynamicParameterizedType.PARAMETER_TYPE, new ParameterizedParameterType(enumClass));
+        setParameterValues(parameters);
     }
 
     public String getName() {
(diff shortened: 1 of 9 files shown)


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

mrotteveel:
This change relies on features introduced in Hibernate 5.3 (eg `org.hibernate.type.descriptor.java.spi.JavaTypeDescriptorRegistry`) and breaks on Hibernate 5.2.

vladmihalcea:
Thanks for the hint. Please open an issue for it

jwgmeligmeyling:
Ah interesting... we compile the 5.2 integration against 5.4 obviously..

I'll have a look. What a bummer. Its undoubtedly going to be difficult to support this on both 5.2 and 5.4, in particular without yet another integration module...

vladmihalcea:
We could check if the class is available on classpath and only support this if we find the class.

mrotteveel:
@vladmihalcea I created #176 for this.

PR #127, opened 2019-09-11
Add YearMonthTimestampType

JWGmeligMeyling:
In some legacy schemas it might be required to map a YearMonth to a Timestamp value, rather than the Date value that we currently support. This may for example be chosen for easy range comparisions to other timestamp values without any casting being required.

This pull request adds such a YearMonthTimestampType. While I think YearMonthDateType will generally be the preferred type, this at least should smoothen the transition from a YearMonth modelled as a timestamp to a YearMonth value.

*Note: I've kept the old author tags because I merely changed `DateTypeDescriptor` to `TimestampTypeDescriptor` and the rest is a pure copy.*


vladmihalcea:
Makes sense. I'll integrate this one too. Thanks.

vladmihalcea:
Applied upstream, thanks

PR #128, opened 2019-09-11
Add support for mapping the Java Period object to a PostgreSQL interval type

JWGmeligMeyling:
In PostgreSQL, an interval may comprise years, months, days, hours, minutes and seconds. In Java, an interval is distinguished between a temporal interval (Duration) and a quantity in terms of days, months and years (Period).

For example `age(timestamp, timestamp)` returns an interval between two timestamps in days, months and years. The result of this function should as such be mapped to a Java Period instead of an Duration.

On the contrary, `timestamp - timestamp` returns an interval between two timestamps in days, hours, minutes and seconds, for which the existing Duration-based type is appropiate.


vladmihalcea:
Looks good.

vladmihalcea:
Applied upstream, thanks

PR #129, opened 2019-09-11
Add support for mapping a Guava Range object to a PostgreSQL range type

JWGmeligMeyling:
Even before there was the `PostgreSQLRangeType`, I had my own range type based on Guava's `Range` class. Lately, while adding `ZonedDateTime` support, I stumbled upon some issues in my parser, and while looking for new solution I saw the existing solution in this repository. My solution has since converged to the one already in the repository. However, I have not been able to fully transition to the `PostgreSQLRangeType`: with a lot of code already using the `Range` API, it is not trivial for me to change my logic back to use `com.vladmihalcea.hibernate.type.range.Range` instead.

I'd understand if you're not interested in merging this pull request because of the redundant nature. However, considering the fact that I am sitting on these changes and that others might be interested in it as well, I thought that I would at least put it out there.

If you do decide to merge this, please do note that I have put the classes in the `com.vladmihalcea.hibernate.type.range.guava` package. While this makes sense for this contribution, this may open the door to other Guava specific types, in which case it would probably be better to create a new `com.vladmihalcea.hibernate.type.guava` package instead.

Note that the infinite range is not supported for this implementation (we cannot detect range type), neither is the empty range (Guava does not support it).

### DIFF
--- a/hibernate-types-4/pom.xml
+++ b/hibernate-types-4/pom.xml
@@ -40,6 +40,14 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-ehcache</artifactId>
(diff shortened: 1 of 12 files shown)


vladmihalcea:
Applied upstream, thanks

PR #136, opened 2019-09-29
Typo in readme

luvarqpp:
Fixed single char typo in README.MD,
causing bad rendering of link.


vladmihalcea:
Applied upstream, thanks

Issue #97, opened 2019-05-24
Add support for mapping ARRAY to a java.util.List


bendem:
You can already map a postgresql ARRAY to a java array, but using raw arrays is not exactly idiomatic in java. Would'nt it make sense to allow mapping it to a `Collection` instead?

vladmihalcea:
Yes, it can be done. Send me a Pull Request with the implementation. Looking forward to it.

nstdio:
@vladmihalcea I can take it, can you assign this issue to me?

vladmihalcea:
Done. Thanks for adding your contribution.

bendem:
Awesome, thank indeed @nstdio

vladmihalcea:
Fixed by [this commit](https://github.com/vladmihalcea/hibernate-types/commit/a53a39b5217a44f74095f0b3c24f6be7966c30a9).

PR #139, opened 2019-10-16
Add type that allows mapping a List entity attribute to a PostgreSQL array column type

hdani9307:


hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/basic/PostgreSQLIntegerListTypeTest.java
```diff
@@ -0,0 +1,86 @@
...
+        @Id
+        @GeneratedValue
+        private Long id;
+
+        @NaturalId
+        @Column(length = 15)
+        private String isbn;
+
+        @Type(type = "integer_list")
+        @Column(columnDefinition = "integer_list")
```
vladmihalcea: I doubt this test will run since I don't see how the `integer_list` PostgreSQL column type is created.

vladmihalcea:
I think this should work for any List<T> type, not just for Integers. The T parameter type can e inferred via the `DynamicParameterizedType`. There can be a paremeter which defines the underlying SQL array type.

Take a look at the JSON types to see how the `DynamicParameterizedType` works. If you have time to rework it, I'll take a look at it after it's done.

vladmihalcea:
Closed as no change was provided according to feedback.

hdani9307:
Can you re-open this PR? I implemented the changes that you suggested in the feedback.

vladmihalcea:
@hdani9307 I'm now creating a generic solution that will allow persisting any type of List<?> into an `Object[]`. Thanks for your contribution.

vladmihalcea:
This issue has been fixed by [this issue](https://github.com/vladmihalcea/hibernate-types/issues/97) and will be released in 2.9.0.

Issue #140, opened 2019-10-17
The PostgreSQLRangeType should use +HH:MM for the tstzrange column type 


partip:
When minutes is represented after semi-colon (:) in timezone (hh:mm), getting DateTimeParseException.

Steps to reproduce,

Add below line to method _com.vladmihalcea.hibernate.type.range.RangeTest.zonedDateTimeTest(_) and execute the test case. 

`assertNotNull(Range.zonedDateTimeRange("[2019-03-28 04:03:10.2146+05:30,)"));`

Here is exception message,

`java.time.format.DateTimeParseException: Text '2019-03-28 04:03:10.2146+05:30' could not be parsed at index 10
`

vladmihalcea:
Thanks for the heads up. You need to provide a replicating test case and a fix as well. I'll review it when it's ready.

partip:
Created Pull request #141. Please review

vladmihalcea:
The test case provided fails, so I'm closing the issue since it does not look like there is anything wrong with the current implementation.

PR #141, opened 2019-10-18
The PostgreSQLRangeType should use +HH:MM for the tstzrange column type

partip:
Pull Request for the #140 issue.


vladmihalcea:

vladmihalcea:
The test fails with this error:

`Caused by: org.postgresql.util.PSQLException: ERROR: invalid input syntax for type timestamp with time zone: "2007-12-03 10:15:30.0+01:00+01"`

Therefore, I'm closing the PR and the issue as well since this syntax is not supported.

PR #154, opened 2019-11-23
ObjectMapperJsonSerializer should inspect the inner Collection element type

joshiste:


joshiste:
Currently Maps are not handled. That would need some special code.

vladmihalcea:

vladmihalcea:
Applied upstream, thanks

PR #159, opened 2019-12-09
Add support for mapping java.util.Date ARRAY or List to PostgreSQL DATE[] and TIMESTAMP[]

Ethoytaryn:
Oups head in the cloud =(. Second PR with DateArrayType class, DateArrayTypeDescriptor and Units test.


vladmihalcea:
Applied upstream, thanks.

I added support for both DATE and TIMESTAMP for both Java array and List entity attribute types.

Issue #170, opened 2020-01-14
ListArrayType should match the initial type when populating the List with the array from the DB


marceloverdijk:
When having a entity like:

```
@Entity
@TypeDefs({
        @TypeDef(name = "list-array", typeClass = ListArrayType.class)
})
public class Circuit {

    @Column(name = "previous_names")
    @Type(type = "list-array")
    private List<String> previousNames = new ArrayList<>();

    ..
```

Could the `ListArrayType` add the names to the initiated list instead of replacing the list?

I'm wondering as when the array is null in the database the `previousNames` list will be null in the fetched entity. I'm having 2 feelings here as it makes sense if it's null in the database.

But I expected differently.

Similar as 

```
    @OneToMany(mappedBy = "project")
    private List<Task> tasks = new ArrayList<>();
```
 
will also give an empty list, even if no tasks would be found in the databse.

Maybe 

```
    @Type(type = "list-array")
    private List<String> previousNames = new ArrayList<>();
```

could return an empty list.

and

```
    @Type(type = "list-array")
    private List<String> previousNames;
```

could return null.

What do you think?

vladmihalcea:
Yes, that can also be an option. It can also be made configurable and passed via a Type parameter.

vladmihalcea:
This [test case](https://github.com/vladmihalcea/hibernate-types/commit/1aea30af27d1699b6e525936bd163aedf7d78398) demonstrates that this behavior is already in place.

marceloverdijk:
In my project it is not working, but there is a difference..

In my database the actual values are `null`.

Looking at the test:

```
    @Test
    public void testEmptyArrays() {

        doInJPA(entityManager -> {
            Event nullEvent = new Event();
            nullEvent.setId(0L);
            entityManager.persist(nullEvent);

            Event event = new Event();
            event.setId(1L);
            entityManager.persist(event);
        });

        doInJPA(entityManager -> {
            Event event = entityManager.find(Event.class, 1L);

            assertArrayEquals(new UUID[]{}, event.getSensorIds().toArray());
            assertArrayEquals(new String[]{}, event.getSensorNames().toArray());
            assertArrayEquals(new Integer[]{}, event.getSensorValues().toArray());
            assertArrayEquals(new Long[]{}, event.getSensorLongValues().toArray());
            assertArrayEquals(new SensorState[]{}, event.getSensorStates().toArray());
            assertArrayEquals(new Date[]{}, event.getDateValues().toArray());
[...]

PR #171, opened 2020-01-14
Added test with null database value for list

marceloverdijk:
Failing test for: https://github.com/vladmihalcea/hibernate-types/issues/170


vladmihalcea:
Just change the getter and setter to:

```
public List<String> getSensorNames() {
    if(sensorNames == null) {
        sensorNames = new ArrayList<>();
    }
    return sensorNames;
}

public void setSensorNames(List<String> sensorNames) {
    if(sensorNames == null) {
        sensorNames.clear();
    }
    this.sensorNames = sensorNames;
}
```

PR #177, opened 2020-01-21
GraalVM / Quarkus native build support - Use Thread.currentThread().getContextClassLoader() when calling Class.forName()

marceloverdijk:
`hibernate-types` does not work with Quarkus (and possible also in other environments) as it is using `Class.forName(className)` at several places.

This PR changes this to use `Class.forName(className, false, Thread.currentThread().getContextClassLoader())` instead. 

See [this](https://github.com/quarkusio/quarkus/issues/6625) discussion on the Quarkus project with the rationale behind the change.

I've tried a local snapshot with the changes with Quarkus and Spring Boot and it works.

Note that locally `mvn clean test` gave a couple of failures, but I see these fail on `master` as well.


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

PR #182, opened 2020-02-10
Improve logging of array parameters in BasicBinder

dpolivaev:
Before:
~~~
binding parameter [3] as [ARRAY] - [[Lcom.xxx.yyy.zzz.xxx.api.SampleTriggers;@5c3ef336]
~~~
Now:
~~~
binding parameter [3] as [ARRAY] - [[PURCHASE_CONFIRMATION]::SampleTriggers[]]
~~~

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/array/internal/AbstractArrayTypeDescriptor.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/array/internal/AbstractArrayTypeDescriptor.java
@@ -68,6 +68,15 @@ public abstract class AbstractArrayTypeDescriptor<T>
     public T fromString(String string) {
         return ArrayUtil.fromString(string, arrayObjectClass);
     }
+    
+    @Override
+    public String extractLoggableRepresentation(T value) {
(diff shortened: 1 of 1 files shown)


vladmihalcea:
Thanks for the Pull Request. I'll review it when I have some time.

vladmihalcea:
Thanks for the PR. I noticed that the Hibernate `PrimitiveByteArrayTypeDescriptor` does not truncate the data. More, using the Hibernate logging is not the best way to log statements. You are better off using [datasource-proxy](https://vladmihalcea.com/the-best-way-to-log-jdbc-statements/).

So, even if this PR is merged, it will not work for other logging solutions. Therefore, you should restrict the size of logged messages in the Log framework.

vladmihalcea:
I reopened it because I thought the issue was about truncating the array. I'll integrate it without truncation.

dpolivaev:
The purpose of truncation was to avoid possible megabytes of data logged as nobody can know the array lengths in general.

dpolivaev:
But you are right, it was mainly not about truncation.

vladmihalcea:
Thanks, I merged it.

PR #183, opened 2020-02-11
Add support for double arrays

wlievens:
I don't have access to a system that supports docker right now (I only run Windows Home edition), so I didn't manage to run the unit tests. Hopefully the change doesn't break any tests. I used the `"double precision"` name for typing the double[] as it is done in PostgreSQL.


vladmihalcea:

wlievens:
It seems like the correct type in JDBC is `float8` rather than the postgres-specific

wlievens:
precision` so that should be changed on line 13 in `DoubleArrayTypeDescriptor.java`.

vladmihalcea:
Applied upstream, thanks

PR #187, opened 2020-02-21
The ListArrayTypeDescriptor deepCopy method should not convert a List to an Java array

dimalu85:
I got issues with casting Object[] from cache and List<> from runtime. That fixes solved my problem.
Used 
```
hibernate.cache.region.factory_class: com.hazelcast.hibernate.HazelcastCacheRegionFactory
hibernate_types_version=2.9.3
hibernate_version=5.4.10.Final
```
Please give me more advice how to describe issue for more.


vladmihalcea:
There are tests for EHCache, so you need to use one of those as a template to create a replicating test case that proves the issue.

Once you add a test case, I will investigate the fix.

vladmihalcea:
Fixed.

PR #199, opened 2020-03-08
Add Iso8601MonthType to save the 1-based value instead of the 0-based Enum ordinal for a java.time.Month entity attribute

### LINKED ISSUE #56: Add Iso8601MonthType to save the 1-based value instead of the 0-based Enum ordinal for a java.time.Month entity attribute
The `java.time.Month` Enum saves the ordinal value starting from 0, even if the underlying month values are 1-based. This type should save the 1-based value in the DB as well.

Postremus:
from 1 (January) to 12 (December)
closes #56


vladmihalcea:
Thanks, I merged it.

Issue #211, opened 2020-04-08
Add support for PostgreSQL "infinity" String values


lborie:
Hi !

With PostgreSQL, when you define a `tstzrange`, you can define it like this :
[`["2017-09-15 11:21:00+02",infinity)`](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-INFINITE)

When I try to load an object with that value as a range, we have an exception : 

```
java.time.format.DateTimeParseException: Text 'infinity' could not be parsed at index 0
	at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1949) ~[?:1.8.0_242]
	at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851) ~[?:1.8.0_242]
	at java.time.ZonedDateTime.parse(ZonedDateTime.java:597) ~[?:1.8.0_242]
	at java.time.ZonedDateTime.parse(ZonedDateTime.java:582) ~[?:1.8.0_242]
	at com.vladmihalcea.hibernate.type.range.Range.lambda$parseZonedDateTime$1(Range.java:447)
```

I understand that in the [`Range.java:256`](https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/Range.java#L256) implementation, an infinite bound must be empty.

How do you think I could get around this ?

Thank you for your time !

vladmihalcea:
Most likely, you will have to debug it and add see what's needed to be changed to support this.

I don't have time to investigate it, but if you provide a Pull Request with a replicating test case and a fix, I'll review it.

vladmihalcea:
Fixed.

PR #213, opened 2020-04-08
Add support for PostgreSQL "infinity" String values

lborie:
This Pull Request is linked to the Issue #211 

It proves that `["2007-12-03T10:15:30+01:00",infinity)` is a valid tstzrange in Postgresql, and it permits, by changing the implementation of `Range.java`, to simply accept the `infinity` keyword into a range.


lborie:
I made 2 different implementations. One (the first commit) that should impact all the ranges, and an other one that impacts only the date ranges.

I think the second (and current) implementation is the best one.

vladmihalcea:
Merged after I did some refactoring so that it works with any Range type.

PR #218, opened 2020-04-20
Allow StringArrayType to work on HSQLDB

mjameson-se:
Without this change, I was getting the unsupported operation exception complaining that the originalArray was of type Object[].


vladmihalcea:
Every chance must be accompanied by a test case. I'll check it out once you provide the test case as well.

vladmihalcea:
Applied upstream.

PR #220, opened 2020-04-23
Allow the custom ObjectMapper be used for JSON serialization

lounagen:
When using a custom JSON serializer, by setting the property `hibernate.types.json.serializer` used in `com.vladmihalcea.hibernate.type.util.Configuration`, the default mechanism doesn't use the custom serializer for `Map` types (HashMap, ...).

The current code handles Collection-based types but fallbacks on the default behaviour for other types.
The default behaviour check if the value to serialize implements `Serializable` (and HashMap does), then the underlying Map values aren't handle by the custom serializer because already in the java native `clone()` behaviour.

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/util/ObjectMapperJsonSerializer.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/util/ObjectMapperJsonSerializer.java
@@ -28,6 +28,14 @@ public class ObjectMapperJsonSerializer implements JsonSerializer {
             }
         }
 
+        if (value instanceof Map && !((Map) value).isEmpty()) {
+            Map.Entry firstElement = (Map.Entry) ((Map) value).entrySet().iterator().next();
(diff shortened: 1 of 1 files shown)


vladmihalcea:

lounagen:
> Thanks, I'll review it when I have some time.

No pb, to avoid being blocked, we switch on a custom serializer which doesn't use your `com.vladmihalcea.hibernate.type.json.internal.JacksonUtil` anymore. I've just push a PR to share the analyse with you but we aren't expecting any urgent update.

I tested this code on a single local unittest under jvm 8, as the `mvn clean install` of the project failed due to expected jvm 6 that i haven't install.

Thanks for your project, really useful!

vladmihalcea:
No worries. I'll make surr sll tests pass when integrating it.

vladmihalcea:
Add some modifications, test cases, and apply it upstream.

vladmihalcea:

lounagen:
> Thanks, I'll review it when I have some time.

@vladmihalcea i think this last comment of you is for the PR #230 opened by @joshiste

PR #225, opened 2020-04-28
Allow JSON Types to handle parameterized entity attribute type variables

charllie:
Handle TypeVariable type

For instance:

```java
@Entity
@TypeDef(name = ColumnTypes.JSON_BINARY, typeClass = JsonBinaryType.class)
@DiscriminatorColumn(name = Columns.TYPE)
public class MyClass<T> {
  @Type(type = ColumnTypes.JSON_BINARY)
  @Column(columnDefinition = ColumnTypes.JSON_BINARY)
  private T something;
}
```

### DIFF
--- a/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonTypeDescriptor.java
+++ b/hibernate-types-43/src/main/java/com/vladmihalcea/hibernate/type/json/internal/JsonTypeDescriptor.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
 import java.sql.Blob;
 import java.sql.SQLException;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


vladmihalcea:

charllie:
> Thanks, I'll review it when I have some time.

No problem, thanks for your library BTW 👍

vladmihalcea:
@charllie You're welcome. Stay tuned for the 2.10 release.

vladmihalcea:
[This test case](https://github.com/vladmihalcea/hibernate-types/commit/d392ef14dd75fec4020b6ea5c61a4d0f1d811f1a) proves that everything works fine even without your change.

If you provide a replicating test case that proves there's a need for this change, I will reopen the issue.

charllie:
Thanks for your reply.

Indeed, this test case is working. I did a project with a failing scenario but I am not able to reproduce it with your test case model.

If you have some times, you can take a look at https://github.com/charllie/hibernate-types-52-spring-boot (failing in BookStarter class)

vladmihalcea:
Applied upstream. Thanks.

PR #230, opened 2020-05-13
Use non-null entries for guessing the content type of maps

joshiste:
It may happen that a map contains keys associated with null values, when these maps are cloned
a NullPointerException is thrown. This was bug introduced with #220. We fix this by only
considering non-null values in the map to guess the content-type for the cloning of
non-serializable maps.

/cc @lounagen


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

Issue #244, opened 2020-06-12
Guava range mapping does not support infinite bounds


azell:
If I create a Guava Range instance like the following:

```
Range.atLeast(LocalDate.of(1934, Month.JUNE, 9))
```

then saving it will generate an exception.  The offending code is in
https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java#L364

According to https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Range.html

the `lowerBoundType` and `upperBoundType` methods will throw an **IllegalStateException** if the corresponding `hasLowerBound` and `hasUpperBound` methods return **false**.

So if the intent is to support lower or higher infinite bounds, then the calls to `lowerBoundType` and `upperBoundType` should have guards.

azell:
And I think there is a copy and paste error at https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java#L173

Should replace `upper` on lines 173 and 174 with `lower`.

vladmihalcea:
Thanks for the analysis. Send me a Pull Request with a replicating test case and a fix to these issues, and I'll review it.

azell:
@vladmihalcea https://github.com/vladmihalcea/hibernate-types/pull/245

vladmihalcea:
Fixed.

PR #245, opened 2020-06-14
Guava Range Type improvements #244

azell:
Various fixes for mapping Guava Range type as detailed in https://github.com/vladmihalcea/hibernate-types/issues/244

### DIFF
--- a/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
+++ b/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
@@ -137,8 +137,8 @@ public class PostgreSQLGuavaRangeType extends ImmutableType<Range> {
                 Ranges.lessThan(upper);
         } else if (upperStr.length() == 0) {
             return lowerBound == BoundType.CLOSED ?
-                Ranges.atLeast(upper) :
-                Ranges.greaterThan(upper);
+                Ranges.atLeast(lower) :
+                Ranges.greaterThan(lower);
         } else {
             return Ranges.range(lower, lowerBound, upper, upperBound);
         }
--- a/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
+++ b/hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
@@ -165,7 +165,6 @@ public class PostgreSQLGuavaRangeType extends ImmutableType<Range> {
      *     Range<Integer> halfOpen = Range.integerRange("(-1,1]");
      *     Range<Integer> open = Range.integerRange("(1,2)");
      *     Range<Integer> leftUnbounded = Range.integerRange("(,10)");
-     *     Range<Integer> unbounded = Range.integerRange("(,)");
      * }</pre>
      *
      * @param range The range string, for example {@literal "[5,7]"}.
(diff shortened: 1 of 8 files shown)


vladmihalcea:

hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
```diff
@@ -81,7 +81,13 @@ protected void set(PreparedStatement st, Range range, int index, SessionImplemen
     }
 
     private static String determineRangeType(Range<?> range) {
-        Object anyEndpoint = range.hasLowerBound() ? range.lowerEndpoint() : range.upperEndpoint();
+        Object anyEndpoint = range.hasLowerBound() ? range.lowerEndpoint() :
+                             range.hasUpperBound() ? range.upperEndpoint() : null;
+
+        if (anyEndpoint == null) {
+            throw new IllegalArgumentException("Cannot find bound type");
```
jwgmeligmeyling, line 88: There might not be a range type here, but this case should probably just return `Range.all()`. The problem with `Range.all()` is that - although we can read it - we cannot write it, because we have no bound to infer the type (and thus the PostgreSQL type for the `PGObject` from). For that it would be nice if the type could be specified as a parameter of the `PostgreSQLGuavaRangeType` instead. Then `Range.all()` can just be mapped fine and we don't have to throw this dubious error.

hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
```diff
@@ -121,14 +127,18 @@ private static String determineRangeType(Range<?> range) {
...
+
         if (lowerStr.length() == 0) {
             return upperBound == BoundType.CLOSED ?
                 Ranges.atMost(upper) :
                 Ranges.lessThan(upper);
         } else if (upperStr.length() == 0) {
             return lowerBound == BoundType.CLOSED ?
-                Ranges.atLeast(upper) :
-                Ranges.greaterThan(upper);
+                Ranges.atLeast(lower) :
```
jwgmeligmeyling, line 140: This fix is actually necessary for the proper working of the type, I have it applied locally as well (just hadn't the time to push it upstream yet).

Not sure about the other change yet. We don't "fix" the support but just throw another type of error. I think we can do more.

hibernate-types-4/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
```diff
@@ -211,12 +221,11 @@ public Long apply(String s) {
     public String asString(Range range) {
         StringBuilder sb = new StringBuilder();
 
-
-        sb.append(range.lowerBoundType() == BoundType.CLOSED ? '[' : '(')
+        sb.append(range.hasLowerBound() && range.lowerBoundType() == BoundType.CLOSED ? '[' : '(')
```
jwgmeligmeyling, line 224: Same for this change. This fix is actually necessary for the proper working of the type, I have it applied locally as well (just hadn't the time to push it upstream yet).

jwgmeligmeyling:
While at it, please also update the `LOCAL_DATE_TIME`  date time formatters, because I've observed various versions of PostgreSQL returning fractions of varying precision for the nanoseconds;

```
 private static final DateTimeFormatter LOCAL_DATE_TIME = new DateTimeFormatterBuilder()
    .appendPattern("yyyy-MM-dd HH:mm:ss")
    .optionalStart() 
   .appendPattern(".")
   .appendFraction(ChronoField.NANO_OF_SECOND, 1, 6, false) 
   .optionalEnd() 
   .toFormatter();
````

vladmihalcea:
@jwgmeligmeyling You could send me a PR according to your findings. Thanks.

jwgmeligmeyling:
@vladmihalcea Sure :)

vladmihalcea:
Fixed.

@jwgmeligmeyling If you have any improvement, please send me a new PR.

PR #266, opened 2020-08-20
Fixed pattern for DateTimeFormatter

gagangoel1994:
At present, the DateTimeFormatter in the range class only works if the time does not contain anything beyond the second precision of it contains all the six digits. 

So let's say if I want to store the time "2020-01-01 01:01:01.111"(up to the millisecond level) within a tsrange, deserialization to the Range object fails. Also, Postgres seems to truncate the trailing zeroes in a given time making all the combinations possible.

Sample code to recreate the issue -

```
public static void main(String []args){
    System.out.println( DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss[.SSSSSS]").parse("2020-01-01 01:01:01.111"));
}
```

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/Range.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/Range.java
@@ -40,7 +40,7 @@ public final class Range<T extends Comparable> implements Serializable {
 
     public static final String INFINITY = "infinity";
 
-    private static final DateTimeFormatter LOCAL_DATE_TIME = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss[.SSSSSS]");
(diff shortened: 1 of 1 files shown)


vladmihalcea:
So, this is needed only to allow the creation of the `Range`.

gagangoel1994:
Postgres supports milliseconds, We can easily insert a value like **'[2020-01-01 01:01:01.111,)'** which is why I feel we should allow this. 

Also, one more thing that happens is if we insert a value like  **'[2020-01-01 01:01:01.111000,)'**, Postgres stores it as  **'[2020-01-01 01:01:01.111,)'**. This creates an issue while deserialization, even if we were passing time till the microsecond level. As you can imagine, because of this we can potentially end up having any number of digits(0-6) for microseconds in the stored Postgres range.

vladmihalcea:
Thanks for investigating it. This issue has been fixed by #267 and will be released in version 2.10.

Issue #262, opened 2020-08-03
Add MonthDayType to map a java.time.MonthDay to either a DATE or an INTEGER column type


vladmihalcea:
Just like `YearMonth`, there should be a Hibernate Type to map a `java.time.MonthDay` to either a DATE or an INTEGER column type

wimdeblauwe:
Is this issue being worked on? I tried to implement it, by basically copy/pasting from YearMonthType, but I can't generate a JAR to test it as I don't have Java 6. Are you interested in a patch file of what I have ?

vladmihalcea:
I'm not working on any issue at the moment since I'm busy with other things. If you have a proof of concept, it would be great to send a PR with it. I'll review it when I have some time, although that might take a couple of weeks or months.

vladmihalcea:
Fixed by #271

PR #271, opened 2020-10-03
Fix issue #262

mladensavic94:
Added mapping for java.time.MonthDay to DATE and INTEGER column types.
@vladmihalcea  please review when possible.


vladmihalcea:
Thanks for the PR. I'll review it when I have some time.

vladmihalcea:
Applied upstream, thanks for your valuable contribution.

Issue #243, opened 2020-06-10
The NotSerializableException is thrown when mapping a Map<String, List<>> entity attribute to a JSON column


z0mb1ek:
Hi. I get exception when try to save entity with field:
```
@Type(type = "jsonb")
@Column(columnDefinition = "jsonb")
private Map<String, List<AudioAttributes>> maxAudioBitratePerOthers;
```

where:

```
@AllArgsConstructor
@NoArgsConstructor
@Builder
@Setter
@Getter
public class AudioAttributes {

    private String name;

    private int bitrate;

}
```

Exception is:

```
Caused by: org.springframework.orm.jpa.JpaSystemException: could not serialize; nested exception is org.hibernate.type.SerializationException: could not serialize
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.convertHibernateAccessException(HibernateJpaDialect.java:353)
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.translateExceptionIfPossible(HibernateJpaDialect.java:255)
	at org.springframework.orm.jpa.AbstractEntityManagerFactoryBean.translateExceptionIfPossible(AbstractEntityManagerFactoryBean.java:528)
	at org.springframework.dao.support.ChainedPersistenceExceptionTranslator.translateExceptionIfPossible(ChainedPersistenceExceptionTranslator.java:61)
	at org.springframework.dao.support.DataAccessUtils.translateIfNecessary(DataAccessUtils.java:242)
	at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:153)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$CrudMethodMetadataPopulatingMethodInterceptor.invoke(CrudMethodMetadataPostProcessor.java:178)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:95)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
[...]

PR #274, opened 2020-10-11
Reproducer for vladmihalcea/hibernate-types#243

pavelgordon:
Hey! 
Based on [issue 243](https://github.com/vladmihalcea/hibernate-types/issues/243), here is a provided test which reproduced the problem.


vladmihalcea:
Thanks for providing me a test case. I will be quite busy in the upcoming 3 months, so I think I'll investigate it in the second half of December or the beginning of January.

pavelgordon:
I decided to implement the fix for it by myself, using the advice you have me in issue discussion. Now tests are green again.
Please review this PR when you will have time.

vladmihalcea:
Great. Thanks for taking the time to fix it. I'll review it when I have some time.

vladmihalcea:
Applied upstream, thanks

Issue #215, opened 2020-04-12
Expose the Hibernate Type for Range elements


jwgmeligmeyling:
The generic type contains information about the element type. In Hibernate 4.3+ this is available through `DynamicParameterizedType.XPROPERTY`. 

This allows for determining the element type, which is useful for example for determining the result type of custom functions, analog to #214.

vladmihalcea:
Fixed.

PR #285, opened 2020-12-19
Support PostgreSQL empty & inf range and OffsetDateTime for Guava ranges

jwgmeligmeyling:
After the `PostgreSQLGuavaRangeType` is now aware of its element type
through the `DynamicParameterizedType` (since #215), the bounds do not
have to be deduced from the provided range anymore. As a result
unbounded ranges like the empty range or the infinite range can be
properly mapped. As a fallback we still use the old behaviour, deducing
the element type from the bound.

For the infinite range nothing special happens. PostgreSQL's `(,)` now
maps to `Range.all()` if the element type is known.

Guava is somewhat opinionated about "empty" ranges. While ranges are
allowed to be empty (there even exists `Range#isEmpty`), a Guava range
is defined by its bounds so for the empty range bounds are still
mandatory (https://github.com/google/guava/issues/2400). PostgreSQL on the other hand normalizes any empty range to an
"empty" literal. As a result bounds will always be lost when persisting
empty ranges in PostgreSQL. I.e. in the previous implementation `[0,0)`
would be persisted fine, but then fail upon a consecutive fetch because
the bounds would have been lost in normalization.

In order to work around this limitation, empty ranges are now mapped to
the empty range with the minimal element value as placeholder bound. At
least always an empty range is returned, but users should note that the
actual bounds are lost during this normalization.

As a second improvement, `tstzrange` now maps to an `OffsetDateTime` by
default, unless a `ZonedDateTime` element is specified through the
`DynamicParameterizedType`. `OffsetDateTime` is more appropiate than
`ZonedDateTime` here, because PostgreSQL only stores the offset and not the
actual timezone. This is also analog to JPA 2.2 which also does
support only `OffsetDateTime` and not `ZonedDateTime`.
When a `ZonedDateTime` is requested, we fall back
to the old behaviour (guessing the timezone based on the offset).

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/range/guava/PostgreSQLGuavaRangeType.java
@@ -21,6 +21,7 @@ import java.sql.Types;
 import java.time.Duration;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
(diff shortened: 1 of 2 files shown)


Tomasvdr198 (approved):

vladmihalcea:
@jwgmeligmeyling Thanks. I'll review it when I have some time.

vladmihalcea:
Applied upstream, thanks for your contribution.

vladmihalcea:
Fixes https://github.com/vladmihalcea/hibernate-types/issues/287

Issue #302, opened 2021-02-27
Use parameter delimiter to join string


sagar-rout:
In StringUtils.java, we are taking delimiter but we are not using it. 

```
public static String join(CharSequence delimiter, CharSequence... elements) {
        StringBuilder builder = new StringBuilder();

        boolean first = true;

        for (CharSequence element : elements) {
            if(first) {
                first = false;
            } else {
                builder.append(LINE_SEPARATOR);
            }
            builder.append(element);
        }

        return builder.toString();
    }
```

sagar-rout:
Hi, I raised PR for the same. Thanks a lot for this awesome project. Please have a look when you have sometime. 

https://github.com/vladmihalcea/hibernate-types/pull/303 

Cheers !

vladmihalcea:
Thanks for the PR. I'll review it when I have some time.

vladmihalcea:
Thanks, I merged it.

PR #303, opened 2021-02-27
Use parameter delimiter to join string and remove hardcoding

sagar-rout:
Use parameter delimiter to join string and remove hardcoding of LINE_SEPRATOR

https://github.com/vladmihalcea/hibernate-types/issues/302#issue-817856475


vladmihalcea:
Applied upstream, thanks

PR #308, opened 2021-03-23
Add fractional support to PostgreSQLIntervalType

lion7:


vladmihalcea:
Applied upstream, thanks

PR #319, opened 2021-05-18
Update JsonBytesSqlTypeDescriptor.java

jbpayne:
I only ran into NPE in this one spot, but it could potentially be other methods and other hibernate versions.


vladmihalcea:
Thanks for spotting it. I'll add it to all modules and run the tests when I have some time.

vladmihalcea:
Applied upstream, thanks

Issue #329, opened 2021-07-05
Add support for half-hour time zones in ZonedDateTime


adeebin:
You had given a great and to the point implementation of range types. I am pretty sure that 1000s may be using this.
I have come across an issue, related to parsing of zoned date-time after saving. Please review whether it's worth looking into.

I have initialised an entity with a zoned timestamp range using

    Range.closedInfinite(ZonedDateTime.now(ZoneOffset.UTC));

On first save (insert), an entry is made in postgres database with time range as 

    id=1, [2021-07-01 17:48:09.97322+05:30, )

Due to some custom requirements, I am loading the object using a native query
```
SELECT 	range_id,	lower(zoned_timerange) as from_time, upper(zoned_timerange) as till_time
FROM	z_buddy where <some condition>
```

The lower and upper are read as timestamptz and is used to update and create entities
Existing:	Range.closedOpen(current.lower(), ZonedDateTime.now(ZoneOffset.UTC)); // id = 1, [2021-07-01 17:48:09.97322+05:30, 2021-07-05 18:47:14.455061+05:30)
New:	Range.closedInfinite(ZonedDateTime.now(ZoneOffset.UTC)); // id is null, [2021-07-05 18:47:14.455061+05:30, )
I.e., on second, save, I close the first record and attempt to create a copy with a new starting range

In my observation, spring tries to load the entity after save, and during that, a parse exception is being thrown when we parse the string according to ZONE_DATE_TIME .
If my observation is right, though the string is in "yyyy-MM-dd HH:mm:ss[.nnnnnn]Z", the matching fails.

As we are suppressing the original exception; we are getting exception on default parse logic of ZonedDateTime
```
private static Function<String, ZonedDateTime> parseZonedDateTime() {
        return s -> {
            try {
                return ZonedDateTime.parse(s, ZONE_DATE_TIME);
            } catch (DateTimeParseException e) {
                return ZonedDateTime.parse(s);
            }
        };
    }
```

I could not find a pattern that works (using DateTimeFormatter); but when I have copy-pasted your code and given a custom parse function, for matching 2021-07-05 18:47:14.455061+05:30, it started working.

In my observation, the definition of formatting at postgres is open and hence, please review whether we can make this pattern configurable, or extendable.

Entity:
--------
```
@Entity(name = "Buddy")
@Table(name = "z_buddy")
@TypeDef(typeClass = PostgreSQLRangeType.class, defaultForType = Range.class)
public class ZonedBuddy {
[...]

PR #331, opened 2021-07-06
Add support for half-hour time zones in ZonedDateTime

adeebin:
- "X" pattern was creating issue, as it was not matching +05:30, it was only matching +05
- Offset "+HH:mm" pattern is suggested as it matches both +05:30 and +05
- Warning: Offset "+HH:mm" pattern will not match +0530, hope PostgesSQL will not return so

Refer: https://github.com/vladmihalcea/hibernate-types/issues/329


vladmihalcea:

vladmihalcea:
Thanks, I merged it.

Issue #368, opened 2021-11-25
protobuf support


nikoncode:
We are using protobuf to store data. So are you planning to add support for protobuf to be (de)serialized into jsonb? It's easy to implement.

vladmihalcea:
At the moment, I don't have time to work on OSS in my spare time. So, there are two ways this feature could be implemented:

1. You can implement it yourself and provide a Pull Request. Make sure to add it to all modules and provide test cases.
2. Or, I could do it via consulting.

nikoncode:
@vladmihalcea I will implement my idea inside my project and if it will work will try to return code into `hibernate-types`.

just wondering does `hibernate-types` support hibernate 5.6.x?

vladmihalcea:
> I will implement my idea inside my project and if it will work will try to return code into hibernate-types.

You can fork this repository, implement the idea, test it on your project and send it as a Pull Request.

> just wondering does hibernate-types support hibernate 5.6.x?

Most likely, yes. Otherwise, [this issue](https://github.com/vladmihalcea/hibernate-types/issues/367) would have shown a problem, rather than just a Hibernate deprecation warning.

nikoncode:
@vladmihalcea it's quite hard to start working because of missing contribution guide. 

Usually when I am trying to implement something for opensource I am starting from clean install, but in this case it's not really possible because I have all tests failed.

vladmihalcea:
Of course, it's possible. Otherwise, there wouldn't have been any contributions.

Your tests are failing because the DB settings don't match your environment. 

No one had the time to implement a contribution guide or set up Testcontainers to ease the ramp-up process. You can start with that. Adding Testcontainers is an easy task and will fix your issue.

peyerroger:
Hi @nikoncode and @vladmihalcea .

We're also interested in storing protobuf as jsonb columns in postgres. I'm happy to help out here, but I'm a bit lost at this point. So any advice how to tackle this issue would be greatly appreciated.

@nikoncode Have you already a somewhat working version of this feature? If yes, what's missing to push it over the finish line?

Thanks for your answers,
Roger

vladmihalcea:
@peyerroger There are plenty of test cases in the repository you can run to see how the existing Types are working.

PR #369, opened 2021-11-26
test containers support for tests #368

nikoncode:
@vladmihalcea please review my changes for 52 before I will replicate it for all packages.

I have 1 PG test failed, but it also failed without my changes (using docker-compose). (most likely it's environmental issue with timezones)

```
com.vladmihalcea.hibernate.type.range.guava.PostgreSQLGuavaRangeTypeTest#test
java.lang.AssertionError: 
Expected :[2007-05-03T10:15:30+01:00..2008-05-03T10:15:30+01:00]
Actual   :[2007-05-03T11:15:30+02:00..2008-05-03T11:15:30+02:00]
```

### DIFF
--- a/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/DataSourceProvider.java
+++ b/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/DataSourceProvider.java
@@ -12,20 +12,5 @@ public interface DataSourceProvider {
 
     DataSource dataSource();
 
-    Class<? extends DataSource> dataSourceClassName();
-
-    Properties dataSourceProperties();
-
-    String url();
-
-    String username();
-
-    String password();
-
     Database database();
-
-    enum IdentifierStrategy {
-        IDENTITY,
-        SEQUENCE
-    }
 }
--- a/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/H2DataSourceProvider.java
+++ b/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/H2DataSourceProvider.java
@@ -1,10 +1,8 @@
 package com.vladmihalcea.hibernate.util.providers;
 
 import org.h2.jdbcx.JdbcConnectionPool;
-import org.h2.jdbcx.JdbcDataSource;
 
 import javax.sql.DataSource;
-import java.util.Properties;
 
 /**
  * @author Vlad Mihalcea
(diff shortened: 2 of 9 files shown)


vladmihalcea:
Thanks, I'll review it when I have some time. Unfortunately, I don't have too much time at the moment.

hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/DataSourceProvider.java
```diff
@@ -12,20 +12,5 @@
 
     DataSource dataSource();
```
vladmihalcea, line 14: I wouldn't delete those methods as they could be useful. These are JDBC client-specific settings, so they should not interfere with this feature.

hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/H2DataSourceProvider.java
```diff
@@ -19,41 +17,12 @@ public String hibernateDialect() {
     @Override
     public DataSource dataSource() {
         return JdbcConnectionPool.create(
-            url(),
```
vladmihalcea, line 22: It's better to keep the existing setting as some of the existing committers might use them.

hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/util/providers/MySQLDataSourceProvider.java
```diff
@@ -1,73 +1,23 @@
...
 
 /**
  * @author Vlad Mihalcea
  */
 public class MySQLDataSourceProvider implements DataSourceProvider {
+    private static final MySQLContainer<?> MYSQL_CONTAINER;
 
-    private boolean rewriteBatchedStatements = true;
+    static {
+        MYSQL_CONTAINER = new MySQLContainer<>("mysql:5.7")
```
vladmihalcea, line 15: I wouldn't start the container unless we cannot find the DB using the default settings. So, I'd use lazy initialization for it at bootstrap time. If we get a connection failure, then we bootstrap the Testcontainers DB.

nikoncode:
@vladmihalcea if I will do initialization in the constructor, it will start/stop container every test class, I don't believe we expect that. Especially when we are running the whole suit.

Now, I see two options:
* since I know the default url, username, password I can implement a check on "static" level before running container. If I can connect using default credentials, then skip test containers. Otherwise, it will spin the container once.
* taking into account that you don't want test containers to be default behaviour maybe it's better for me to just actualize docker-compose and save efforts for the next contribution?

vladmihalcea:
The docker-compose option is also a good option.

vladmihalcea:
Support for Testcontainers was added by [this commit](https://github.com/vladmihalcea/hibernate-types/commit/a571dbe03c04991361ace4eaeaeeee2352e81921). Thanks for your PR.

PR #371, opened 2021-12-09
Support java.sql.Date[] to java.time.LocalDate[] retrieval

AndrewLazarus:
Storing LocalDate[] to a (Postgres, at least) DB
worked with existing code. Read did not.

Also extended attempt to copy other types of
received arrays to try harder before failing.

[Notes: Java 8 is long gone from my build system, making it difficult
to build and use this project in IntelliJ, at least for me. I created
a jar by hand, built with no errors, and it worked correctly
for my use case, which is given in the title. But the JUnit
tests have _not_ been run on the modified code.]


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

PR #420, opened 2022-04-07
remove extra 0 on ehcache dependency

sanderxavalon:
Please review my change on POM.xml, I only change the version number of ehcache.

I tried to run the latest test a few minutes ago and maven keeps showing me errors.
It disturbing for a while but I found the version number was not match with Maven Repository, and it's good after I remove that.


sanderxavalon:
It works fine today, so weird.
I'm sorry cause you any inconvenience.

vladmihalcea:
No worries.

PR #425, opened 2022-04-11
Add mssql2019 to docker compose file.

sanderxavalon:
Hi Vlad,

Good day, please kindly review my pull request, I add MSSQL in docker-compose and some initial script.
I try to make same trick like MySQL to create a "docker-entrypoint-initdb.d", but seems MSSQL doesn't work that way.
And I also update all the MSSQL connection passwords, I tried to use your password but MSSQL image said it's not strong enough.

I hope this will help, thanks.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -31,3 +31,22 @@ services:
       options:
         max-size: "50m"
         max-file: "1"
+
+  sqlserver-hibernate:
+    image: mcr.microsoft.com/mssql/server:2019-GA-ubuntu-16.04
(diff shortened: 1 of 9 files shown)


vladmihalcea:

vladmihalcea:
This is going to require me to make a lot of changes on my end since I'm using the same setup on multiple projects. Therefore, I'll have to think of a way to expose the URL, credentials as external variables which others can customize without any code change, using Maven profiles or something similar.

vladmihalcea:
This is no longer needed since now the Project uses Testcontainers for all databases.

PR #429, opened 2022-04-15
Change batch sequence query

### LINKED ISSUE #428: Batch sequence generation broken on H2 2.x
Due to https://github.com/h2database/h2database/issues/3483 batch sequence generation is broken on versions 2.x of H2.

There are two possible fixes:

We could special case the H2 dialect and use this query

```sql
SELECT NEXT VALUE FOR sequenceName
  FROM SYSTEM_RANGE(1, ?);
```

Or we could simplify / change the default query to

```sql
WITH RECURSIVE T(N) AS (
  SELECT 1
   UNION ALL
  SELECT N + 1
    FROM T
   WHERE N < ?
)
SELECT NEXT VALUE FOR sequenceName
  FROM T;
```

marschall:
Change batch sequence query to be simpler
and work with H2 2.x.

Fixes #428


marschall:
I tried to run the tests. I used the docker compose file to start mysql and postgres but did not find a way to exclude other tests like Oracle.

PR #451, opened 2022-07-05
Update Guava version to version 31.1

loliman:
Guava should be updated to 31.1 because of a recent vulnerability in 29.0. 

Compiles and works just fine with 31.1.


kayman-mk:
Would be great to get this fixed as company guidelines @Hapag-Lloyd prohibit using a library with common vulnerabilities.

Link to the CVE: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908

vladmihalcea:
@kayman-mk I could integrate this issue, but it's actually useless because Guava is an optional dependency, so Hibernate Types just uses whatever Guava version you configured in your project.

Therefore, you don't risk anything no matter whether I integrate this or not.

vladmihalcea:
Applied upstream, thanks

Issue #453, opened 2022-07-11
Avoid the double registration of DoubleArrayType in HibernateTypesContributor for PostgreSQL82Dialect


OctoOsmo:

vladmihalcea:
Fixed by #454

PR #454, opened 2022-07-11
Double registration of DoubleArrayType

OctoOsmo:
Fix of Double registration of DoubleArrayType in HibernateTypesContributor f…or PostgreSQL82Dialect #453


vladmihalcea:
Good catch. I'll integrate it when I get back to work.

PR #455, opened 2022-07-11
Add very first money types handling

### LINKED ISSUE #450: Add support for JavaMoney (JSR 354) with the MonetaryAmountType
In response to https://github.com/JavaMoney/jsr354-ri/issues/185#issuecomment-1146796277 maybe is a good idea to create issue here.

I think handling JavaMoney could be very useful.

This feature is in Jadira already (https://github.com/JadiraOrg/jadira/tree/master/usertype.core/src/main/java/org/jadira/usertype/moneyandcurrency/moneta) but seems it's looks like a little bit stale.

piotrooo:
Resolves #450

### DIFF
--- a/hibernate-types-4/pom.xml
+++ b/hibernate-types-4/pom.xml
@@ -48,6 +48,14 @@
         </dependency>
 
         <dependency>
+            <groupId>org.javamoney</groupId>
+            <artifactId>moneta</artifactId>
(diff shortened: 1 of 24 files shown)


vladmihalcea:

vladmihalcea:
Thanks, I merged it.

PR #456, opened 2022-07-11
Amazing Hibernate Repository

ecopsdee:
This is the update on the Amazing Hibernate Repository.

### DIFF
--- a/hibernate-types-4/pom.xml
+++ b/hibernate-types-4/pom.xml
@@ -65,11 +65,11 @@
     </dependencies>
 
     <properties>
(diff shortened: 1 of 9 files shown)


vladmihalcea:
Thanks. I'll review it when I get back to the office

vladmihalcea:
Thanks, but this PR has so many issues:

- you set the Hibernate version to 6 on all modules, which is so wrong since the foal of each module is to work with a designated Hibernate version range
- you added additional repo references that were not needed
- you didn't add the actual Repository to all the modules
- you did not add any Spring-related tests

Therefore, I'm going to close it, and I'll do it myself instead.

PR #458, opened 2022-07-15
Fix backward compatibility with OffsetDateTime stored as double

### LINKED ISSUE #457: Allow OffsetDateTime to be parsed from a Double value
In version 2.14.1, changes to handlling of OffsetDateTime type values was introduced to persist timezone. Earlier, OffsetDateTime type values were converted to a double type for storage and converted back to OffsetDateTime type on retrieval. 

**The earlier versions also were able to retrieve values stored as OffsetDateTime strings**. Starting v2.14.1, during retrieval, the double type values are not parsed by the library and throw a DateTimeParseException.

```
exception: java.time.format.DateTimeParseException: Text '1657220671.920877' could not be parsed at index 0
	... 5 common frames omitted
Wrapped by: com.fasterxml.jackson.data.JsonMappingException: Text '1657220671.920877' could not be parsed at index 0 (through reference chain: com.company.org.project.service.api.v1_0.model.Schedule["triggerTime"])
	... 5 common frames omitted
Wrapped by: java.lang.IllegalArgumentException: The given string value: {"triggerTime": 1657220671.920877} cannot be transformed to Json object
	... 5 common frames omitted
Wrapped by: org.hibernate.HibernateException: java.lang.IllegalArgumentException: The given string value: {"triggerTime": 1657220671.920877} cannot be transformed to Json object
	... 5 common frames omitted
Wrapped by: org.springframework.orm.jpa.JpaSystemException: java.lang.IllegalArgumentException: The given string value: {"triggerTime": 1657220671.920877} cannot be transformed to Json object; nested exception is org.hibernate.HibernateException: java.lang.IllegalArgumentException: The given string value: {"triggerTime": 1657220671.920877} cannot be transformed to Json object
	at...
```

dan-lind:
This should fix #457 
Only supports _reading_ the double format, OffsetDateTime will still be written as dates

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/util/ObjectMapperWrapper.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/util/ObjectMapperWrapper.java
@@ -11,6 +11,9 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.lang.reflect.Type;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeParseException;
+import java.util.Date;
 
 import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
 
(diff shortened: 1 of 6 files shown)


vladmihalcea:

vladmihalcea:
Check out the question I asked on the issue. It's not clear where that double value comes from. Unless this value comes from the standard Hava Date/Time API, this change will not be implemented as it would not be a generic requirement.

vladmihalcea:
Applied upstream, thanks

Issue #463, opened 2022-07-31
MappingException: Type not supported for auditing


msdousti:
## Synopsis

`ImmutableType` and its descendants like `PostgreSQLHStoreType` are not compatible with **Envers**.
For instance, consider the following `@Audited` Entity:

```java
import com.vladmihalcea.hibernate.type.basic.PostgreSQLHStoreType;

// ...
@Entity
@TypeDef(name = "PostgreSQLHStoreType", typeClass = PostgreSQLHStoreType.class)
@Audited
public class Post {
    @Id
    private Long id;

    @Type(type = "PostgreSQLHStoreType")
    @Column(columnDefinition = "hstore")
    private Map<String, String> translations;

    // ...
}
```

It will cause an
```
org.hibernate.MappingException: Type not supported for auditing: com.vladmihalcea.hibernate.type.basic.PostgreSQLHStoreType
````
exception upon building the session factory.

## Test
A full-fledged test along with explanations can be found [here](https://github.com/msdousti/hstore-envers).

## Fix
`ImmutableType` can be modified to implement `BasicType` instead of `Type`, as explained [here](https://github.com/msdousti/hstore-envers#workaround).

## Suggestion
I can open a PR for applying the above change.

vladmihalcea:
Looking forward to your Pull Request that fixes this issue

msdousti:
Thanks a lot for considering my suggestion. I created https://github.com/vladmihalcea/hibernate-types/pull/464 for `hibernate-types-55`. If the changes there are OK, I'll replicate them to other `hibernate-types-*` modules in the same PR.

Incidentally, I noted that the PR also fixes this:
https://stackoverflow.com/q/69313708/459391

vladmihalcea:
I'll integrate it after I apply the same login to the new `ImmutableCompositeType`

vladmihalcea:
Fixed.

PR #464, opened 2022-08-01
Fix Envers compatibility with ImmutableType

msdousti:
Envers expects (simple) audited types to implement the `BasicType` interface. This commit changes The `ImmutableType` to implement `BasicType` rather than `Type`.

The first commit is work in progress: It only applies the change to  `hibernate-types-55` as a template. If the change is OK, it will be applied to other `hibernate-types-*` as well.

Resolves: #463

### DIFF
--- a/hibernate-types-55/pom.xml
+++ b/hibernate-types-55/pom.xml
@@ -25,6 +25,13 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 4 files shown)


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

Issue #465, opened 2022-08-02
MonetaryAmountType throws NullPointerException when reading a null column value


ugrave:
The MonetaryAmountType  use the ResulSet.wasNull to check for null.
Problem is that at least one getter must called before.

From javadoc og `ResultSet.wasNull`:
`Note that you must first call one of the getter methods on a column to try to read its value`

I attached a demo with two use cases.
One read only the Money value there other one reads the hole entity.
When reading only the money value the npe is thrown. Only whan an additional column was read before it works without exception.

[demo.zip](https://github.com/vladmihalcea/hibernate-types/files/9242622/demo.zip)

vladmihalcea:
I don't understand the issue. There is a integration test in repository you could use to add a new test that replicates your issue.

Not only that it's easier to isolate the problem that way, but, after the fix is done, there is already a test to prove the fix.

ugrave:
Hi,

added a test case which reproduce the error. Found out that this always happend if the salery is null.

https://github.com/ugrave/hibernate-types/blob/96d985b2609c2fcec3436f4f666c03569c61e274/hibernate-types-55/src/test/java/com/vladmihalcea/hibernate/type/money/PostgreSQLMonetaryAmountTypeTest.java#L69

ugrave:
A fix could be to check the amaout value if it is not null instead of the `rs.wasNull()`.
Or first read the amount column and the check with `rs.wasNull()`.

vladmihalcea:
Thanks @ugrave for the tests and the investigation. I'll check it out when I have some time.

In the meanwhile, if you want, you could try to provide a Pull Request with a fix proposal.

vladmihalcea:
Fixed.

RoarBjur:
When will 2.17.2 be released with this fix? This is a major problem for my usage of the MonetaryAmountType class

vladmihalcea:
@RoarBjur it depends on how much your company is willing to pay. For free, it's going to be ready when it's ready.

RoarBjur:
@vladmihalcea Thank you for your reply. I am not in a hurry right now, but will contact you if necessary

vladmihalcea:
There is another issue that I plan to add to 2.17.2, so it all depends on when I have the time to get that done.

PR #467, opened 2022-08-03
Fix npe if money is null

ugrave:
Closed gh-465

### DIFF
--- a/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
+++ b/hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
@@ -75,14 +75,14 @@ public class MonetaryAmountType implements CompositeUserType {
 
     @Override
     public Object nullSafeGet(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner) throws HibernateException, SQLException {
-        if (rs.wasNull()) {
-            return null;
-        }
-
         String amountColumnName = names[0];
         String currencyColumnName = names[1];
 
         BigDecimal amount = rs.getBigDecimal(amountColumnName);
+        if(amount == null) {
+            return null;
+        }
+
         String currency = rs.getString(currencyColumnName);
 
         return Money.of(amount, currency);
--- a/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/money/MySQLMonetaryAmountTypeTest.java
+++ b/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/money/MySQLMonetaryAmountTypeTest.java
@@ -12,6 +12,7 @@ import javax.persistence.*;
 import java.math.BigDecimal;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 /**
  * @author Piotr Olaszewski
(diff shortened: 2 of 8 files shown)


vladmihalcea:

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
```diff
@@ -75,14 +75,14 @@ public int hashCode(Object x) throws HibernateException {
...
-
         String amountColumnName = names[0];
         String currencyColumnName = names[1];
 
         BigDecimal amount = rs.getBigDecimal(amountColumnName);
+        if(amount == null) {
+            return null;
+        }
+
         String currency = rs.getString(currencyColumnName);
```
piotrooo, line 87: Null-check should be here. Because creating monetary type without currency make no sense and could rise an exception `Currency Code may not be null`.

I propose to use here:
```
if (rs.wasNull()) {
    return null;
}
```

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
```diff
@@ -75,14 +75,14 @@ public int hashCode(Object x) throws HibernateException {
...
     public Object nullSafeGet(ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner) throws HibernateException, SQLException {
-        if (rs.wasNull()) {
-            return null;
-        }
-
         String amountColumnName = names[0];
         String currencyColumnName = names[1];
 
         BigDecimal amount = rs.getBigDecimal(amountColumnName);
+        if(amount == null) {
```
piotrooo, line 82: Removing this null-check in favor of `rs.wasNull()` (see next comment).

piotrooo (changes requested):
@ugrave great catch!

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
ugrave: Question is should the system thow an exception if only one of them is null or still ignored it an return also null in that case. I`m was not sure about this case and decide to let the exception be thrown.

Is there a diffferend if i use `rs.wasNull()` or check the value of the last read column?
To get the correct result of `rs.wasNull()` i need to read the column before. It also only checks if the last read column is null.

ugrave: Because the Java doc of the `rs.getBigDicimal(x)` says: `if the value is SQL NULL, the value returned is null`

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
vladmihalcea: @piotrooo The `wasNull` method call is useful only when you have primitive values on the Java side, but this is not the case here. We can use `== null` checks.

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
piotrooo: `null` seems to be enough.

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
piotrooo: `null` seems to be enough.

hibernate-types-52/src/main/java/com/vladmihalcea/hibernate/type/money/MonetaryAmountType.java
piotrooo: But we should reconsider strategy. If any of a column has `null` - that should be `null` or exception? Maybe if any value is `null` - `null` should be returned.

vladmihalcea:
Applied upstream, thanks.

PR #474, opened 2022-08-12
Add support for JavaMoney (JSR 354) with the CurrencyType #473

### LINKED ISSUE #473: Add support for the JavaMoney (JSR 354) CurrencyUnit
Hello @vladmihalcea!
I hope you're fine.

I come with another idea about extending types. We can add currency ISO-4217 codes.
It looks like:

```
@Column(name = "currency")
private CurrencyUnit currency;
```

What do you think? If you are interested in, I implemented almost everything what is needed.

Another question, I see you add Money handling for Hibernate versions >5, that is intentionally? Because I make implementation for all of Hibernate versions. If is, maybe pom's should be cleaned and removed Moneta dependency.

If it's ok, I can implement for all Hibernate versions >5.

piotrooo:
Closes #473

### DIFF
--- a/hibernate-types-4/pom.xml
+++ b/hibernate-types-4/pom.xml
@@ -48,15 +48,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.javamoney</groupId>
-            <artifactId>moneta</artifactId>
-            <version>${moneta.version}</version>
-            <type>pom</type>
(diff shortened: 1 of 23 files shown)


vladmihalcea:

vladmihalcea:
Merged

PR #495, opened 2022-10-17
added SetArrayType, SetArrayTypeDescriptor

Aleksey2093:


vladmihalcea:
Fixed.

khaledE:
somehow this completely disappeared, I can't find the SetArrayType in the latest version 
`                <groupId>io.hypersistence</groupId>
                <artifactId>hypersistence-utils-hibernate-55</artifactId>
                <version>3.7.3</version>`

vladmihalcea:
The `ListArrayType` works with both List and Set. Here's an example:

https://github.com/vladmihalcea/hypersistence-utils/blob/master/hypersistence-utils-hibernate-63/src/test/java/io/hypersistence/utils/hibernate/type/array/ListArrayTypeTest.java#L536

khaledE:
oh interesting for some reason I keep getting this error when I do the same as in the example

org.springframework.orm.jpa.JpaSystemException: Could not set field value [[SOME_KEY]] value by reflection
nested exception is org.hibernate.PropertyAccessException: Could not set field value [[SOME_KEY]] value by reflection

vladmihalcea:
Check out that example I sent you as it works like a charm. You can compare them in debug to see where they differ.

khaledE:
yes but that only works in the newer versions like 60 or 62
the change was never merged to hypersistence-utils-hibernate-55
I'll keep using my workaround until we're ready to upgrade

thanks!

vladmihalcea:
@khaledE Then how do you explain that there's also a test case in the 55 module that also works fine with `Set<LocalDateTime>`?

https://github.com/vladmihalcea/hypersistence-utils/blob/master/hypersistence-utils-hibernate-55/src/test/java/io/hypersistence/utils/hibernate/type/array/ListArrayTypeTest.java#L508

PR #502, opened 2022-10-20
Do not need to clone strings with the generic serializer cloner

gzsombor:
as they are immutable.

Profiling my application, I've noticed, that a lot's of time spent in the cloning of strings with SerializationHelper.clone(...) - it took 5%-6% on average, which seemed unnecessary.
 In my entity, the JSON column mapped to a single String field, because the app doesn't need to deal with the actual content of json, just gets the value from PostgreSQL and sends back to the client.
With this change, this cloning doesn't happen anymore.


vladmihalcea:

Issue #507, opened 2022-10-28
Empty ranges cannot be converted back to their initial String representation


nstdio:
Consider testcase:

```java
@Test
public void asStringWithEmptyKeyword() {
    Range<Integer> empty = Range.integerRange("empty");

    assertEquals("empty", empty.asString());
}
```
should pass, but exception is thrown:

```
java.lang.NullPointerException: Cannot invoke "Object.toString()" because "t" is null

	at com.vladmihalcea.hibernate.type.range.Range.lambda$boundToString$3(Range.java:673)
        .....
```

vladmihalcea:
Fixed.

PR #508, opened 2022-10-28
Empty Range cannot be converted to string.

nstdio:
Closes: gh-507


vladmihalcea:
Applied upstream, thanks

PR #550, opened 2023-01-04
add range method tests

whithang:
i added 3 tests to show that the closed method for the upper bound is never correctly setting the bound as closed. it is always saving it as open. however the lower bound is working as expected. 2 of the 3 tests are failing


whithang:
pull request regarding issue [548](https://github.com/vladmihalcea/hypersistence-utils/issues/548 )

PR #584, opened 2023-02-23
Add @Transactional annotation to BaseJpaRepository methods

depzor:
Since `org.springframework.data.jpa.repository.support.SimpleJpaRepository` is marked as
`@Transactional( readOnly = true )` every persist, merge, update, delete action will be automatically marked as read only as well. 
`Caused by: java.sql.SQLException: Connection is read-only. Queries leading to data modification are not allowed`

Unless the outer method is annotated with `@Transactional`.

This PR solves that problem when there is no transactional annotation present in the first place.


vladmihalcea:
That's true. Thanks for the PR.

vladmihalcea:
Applied upstream, thanks

PR #593, opened 2023-03-16
Correct minor typo in HibernateRepository

curena:
The interface `HibernateRepository` contains utility implementations for bulk operations: `updateAllAndFlush`, `mergeAllAndFlush`, and `peristAllAndFlush`. 

There is a typo in the latter, which should be `persistAllAndFlush`. This contribution aims to correct this typo in the least disruptive way possible. 

Please note that this will constitute a breaking change for any consumers making use of the extant method.


vladmihalcea:

PR #595, opened 2023-03-17
Unencapsulate Range utility functions

Hc747:
Unencapsulate Range bounds checking utility functions.

`Range#hasEqualBounds`
`Range#isBoundedOpen`
`Range#isBounded`


vladmihalcea:

piotrooo (changes requested):
Other implementations for Range should be changed as well.

Hc747:
> Other implementations for Range should be changed as well.

Resolved. Didn't realise this implementation was duplicated across modules. Good catch!

vladmihalcea:
Applied upstream, thanks

PR #603, opened 2023-04-01
Make Hypersistence Utils compatible with Hibernate 6.2.0.Final

cigaly:
Minor fixes to implement new/changed method in Hibernate 6.2.0.Final interfaces


vladmihalcea:
Applied upstream, thanks

PR #611, opened 2023-04-16
BatchSequenceGenerator should be able to search the DB sequence in the default schema

eazrael:
Sequences for Batch ID Generation can now be located in the default schema of the database connection without hardcoding the schema name in the annotation.  Implemented for 6.0 and 6.2. Tested with DB2 and PostgreSQL.


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

PR #612, opened 2023-04-16
PostgreSQL cteless batch id generation

eazrael:
Changed batch ID fetch query for PostgreSQL from recursive CTE to simple query, reducing execution time by half. Tested with requests to 1 000 000 ids at once. Base branch is [eazrael:default_schema_support](https://github.com/eazrael/hypersistence-utils/tree/default_schema_support)


vladmihalcea:

vladmihalcea:
Applied upstream, thanks

PR #615, opened 2023-04-19
Save a new entity Book with a simple Map

dyleph:
In this pull request, we want to test a case with saving a new entity with a simple Map.
After saving an entity, we want to check its version in SpringDataJPAHibernateTest1.java


vladmihalcea:
The example is too complicated. You need to remove Spring and Lombok and follow the same test structure used by all the other Type tests.

PR #616, opened 2023-04-19
Save a new entity with a nested Map Map

dyleph:
In this pull request, we want to test a case with saving a new entity with a nested Map<Map>. `Map<Publisher, Map<BookTypes, BookInventory>> bookDetails` 
After saving an entity, we want to check its version in SpringDataJPAHibernateTest2. java

Actually, its version is 2 in this project, while expected is 0.
However, in my own project, the version is 1, while expected is 0.


vladmihalcea:
The example is too complicated. You need to remove Spring and Lombok and follow the same test structure used by all the other Type tests.

PR #619, opened 2023-04-24
An extra UPDATE is triggered when using records

dyleph:


vladmihalcea:
Thanks. It's much better. I'll check it out tonight.

vladmihalcea:
Likewise

vladmihalcea:
The problem comes from the `areEqual` method which finds different internal Map types.

dyleph:
Is this problem from hypersistence-utils?

vladmihalcea:
Yes, but you can overcome it if you use a `JsonNode` or a POJO instead.

I will see how to accommodate a fix when I will get some time.

dyleph:
Hi @vladmihalcea ,
Do we have any updates with this issue?

Regards,
Dy

vladmihalcea:
@dyleph This is a very complex issue, and you are better off using `JsonNode` or type-safe POJOs on the entity attribute that maps the JSON if you want to avoid it.

I'm not sure when I'll have time to fix it since the problem is caused by records that can be contained on any hierarchical level in a `Map` of `Map` structure you're now using.

But, as long as you have some workarounds, you should those instead.

Issue #655, opened 2023-09-20
Range<ZonedDateTime>  produces incorrect results (PostgreSQL)


lbruun:
## Describe the bug

`Range<ZonedDateTime>` is no longer the same after having been persisted and then retrieved.  This produces incorrect is-in-range? results.

This is not so much the library's fault as it is a consequence of how PostgreSQL timestamps travel the wire and are interpreted by the JDBC driver.

## To Reproduce

Consider this:
```java
ZonedDateTime zdt = ZonedDateTime.parse("2007-12-03T09:00:00Z");
Range<ZonedDateTime> zdtr = Range.closedInfinite(zdt);
```

This produces a range like this `['2007-12-03T09:00:00Z', INFINITY)`. 

And we have `zdtr.contains(zdt) => true`  (as expected).

Now save the Range into the database (in a column of type `tztsrange`, of course) and retrieve it again into new variable, let's call it `zdtrRetrieved`. Now the Range looks like this:

`zdtrRetrieved --> ['2007-12-03T10:00:00+01', INFINITY)`.   (depending on your session's timezone, mine is UTC+1).

This shouldn't be a problem, right?  The `zdtr`  and `zdtrRetrieved` are functionally the same, one would think.

Not so: now you have `zdtrRetrieved.contains(zdt) --> false` .. which is certainly unexpected.

THIS BREAKS MOST - IF NOT ALL - USAGE OF THIS TYPE, IMO.

## Expected behavior 

Persisting a `Range<ZonedDateTime>` and retrieving it again should ideally yield the same value. However, this is likely not possible to accomplish with the way JDBC and PostgreSQL works.

## Possible fix

[...]

Issue #288, opened 2021-01-08
Add Support for Instant Range


benstpierre:
Would it make sense to Have a Range of Instants?  We have a range of localdate time but Range of instant is equally useful?

vladmihalcea:
What PostgreSQL column type do you think it's best to represent an `Instant`? I suppose it should be a big interest storing milliseconds since epoch, right?

benstpierre:
The project I am working on now uses Hibernate 5.4.27.Final and when I set HBM2DDL_AUTO to auto it generates timestamp columns for Instant.  We are on Postgres 12.5 with org.hibernate.dialect.PostgreSQL10Dialect.  Seems like timestamp is a good match.

vladmihalcea:
Using Hibernate hbm2ddl is a terrible idea unless you are doing it for the very first incremental script. [Flyway](https://vladmihalcea.com/flyway-database-schema-migrations/) is the way.

Timestamp is not a good match for Instant. LocalDateTime is.

benstpierre:
Yes this is exactly what I am doing... using hbm2ddl for my very first script.  I will then take a dump of the schema and use it for my V1__ script in flyway when the app goes to production.

vladmihalcea:
Hibernate supports mapping a Java type to multiple DB types via SQL descriptors. If you send me a Pull Request that implements this issue, I'll review it.

Kurru:
We found that `timestamp with time zone` best matched Instant. This type converted to a +00:00 offset timestamp. `Timestamp without timezone` matched best with LocalDateTime in our experience

vladmihalcea:
The PostgreSQL `timestamp with timezone` is best mapped to `OffsetDateTime`.

Kurru:
Perhaps, but my answer was that `Instant`  mapped well to `timestamp with time zone` for us. Sure theres an additional +00:00 offset, but thats (seemly) harmless and still represents an instant quite well.

asdm90:
@Kurru 
Have you considered forking the Range and the PostgreSQLRangeType classes yourself and changed ZonedDateTime to Instant instead?

Right now the default deserialization of a tstzrange type in the PostgreSQLRangeType gets mapped to a Range<ZonedDateTime>. If you want to you can convert it manually afterwards in the client code, but changing the current implementation to use Instant would be backwards incompatible for current clients of the Range class.

Kurru:
[...]

Issue #96, opened 2019-05-09
Support for Range of OffsetDateTime


tipenull:
Currently only Range<LocalDateTime> is mapped to tstzrange.
However, the default java type for timestamptz is OffsetDateTime rather than LocalDateTime - which seems adequate since timestamptz doesn't actually carry a time zone, but an offset.
Hence, I'd expect Range<OffsetDateTime> to be mapped to tstzrange.

vladmihalcea:
> However, the default java type for timestamptz is OffsetDateTime rather than LocalDateTime

Where did you get this from? There is no default Java mapping between PostgreSQL column types and Java types.

> since timestamptz doesn't actually carry a time zone, but an offset.

That's also not true. The `timestamptz` column type is a timezone-aware type. When you insert a value into a `timestamptz` column, PostgreSQL will convert the `timestamptz` value to its UTC equivalent.

> Hence, I'd expect Range to be mapped to tstzrange.

Which is already the case for a `Range<ZonedDateTime>`.

Hence, there's no issue.

tipenull:
> Where did you get this from? There is no default Java mapping between PostgreSQL column types and Java types.

It's specified for the PostgreSQL JDBC driver ([https://jdbc.postgresql.org/documentation/head/8-date-time.html](url)). 'Default' was definitely too broad of a phrasing.

> Which is already the case for a Range<ZonedDateTime>.

I meant to specify Range<OffsetDateTime>. Funny that I forgot that of all things.

vladmihalcea:
In that case, you should provide a Pull Request with

nstdio:
@tipenull I think you are right

Here is quote from documentation: `Note that ZonedDateTime, Instant and OffsetTime / TIME WITH TIME ZONE are not supported` and [here](https://stackoverflow.com/questions/30234594/whats-the-difference-between-java-8-zoneddatetime-and-offsetdatetime) is why.

nstdio:
@vladmihalcea @tipenull 

I was trying to figure out how to replace the `ZonedDateTime` to `OffsetDateTime` of it turn's out that it no so easy to do it with backward compatibility. Obviously we can deprecate all things connected with `ZonedDateTime` and provide `OffsetDateTime` equivalents, but if we change mapping for `tstzrange` and return `Range<OffsetDateTime>` instead of `Range<ZonedDateTime>` users can easily experience `ClassCastException`.

After few hours of digging it **seems** there is no way to backward compatible change.

[...]

PR #656, opened 2023-09-21
Support for Instant and OffsetDateTime

lbruun:
## Support for `Range<Instant>` and `Range<OffsetDateTime>`
Support for additional Java range types is added as per the text in bold in table below. 

Supported PostgreSQL types:

| PostgreSQL column type | Java type |
| ----- | -------- |
| `int4range`   | `Range<Integer>` |
| `int8range`   | `Range<Long>` |
| `numrange`   | `Range<BigDecimal>` |
| `tsrange`   | `Range<LocalDateTime>` |
| `tstzrange`   | `Range<ZonedDateTime>`, **`Range<OffsetDateTime>` or `Range<Instant>`** |
| `daterange`   | `Range<LocalDate>` |

The change is fully backwards compatible.

Fixes Issue #288  and Issue #96.

## Support for Instant-based comparison

Comparison functionality which can check for "functionally equivalent" has been added. With the existing "strict comparison" the following two ranges:

```text
     x1 : ['2007-12-03T09:30.00Z',)
     x2 : ['2007-12-03T10:30.00+01:00',)
```
is seen as two completely different ranges and `contains()` produces unexpected results in that the `2007-12-03T09:30.00Z` Instant is not contained in the x2 range. (see Issue #655)

This PR adds new methods which performs Instant-based comparison where applicable, meaning for types `OffsetDateTime` and `ZonedDateTime`. Fixes Issue #655.

I don't see why not any application would want to always do `contains(x, true)` going forward, but I didn't dare break backward compatibility. Hence the new feature is purely opt-in based.  There is probably a slight performance degradation by performing Instant-based comparison instead of strict comparison, but I doubt it is truly measurable.

### DIFF
--- a/hypersistence-utils-hibernate-62/src/main/java/io/hypersistence/utils/hibernate/type/range/PostgreSQLRangeType.java
+++ b/hypersistence-utils-hibernate-62/src/main/java/io/hypersistence/utils/hibernate/type/range/PostgreSQLRangeType.java
@@ -16,23 +16,21 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.ZonedDateTime;
+import java.time.*;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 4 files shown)


vladmihalcea:

vladmihalcea:
@lbruun This PR is just for the 6.2 branch, therefore I cannot integrate it unless the changes are provided for all modules.

PR #658, opened 2023-09-22
Add JFR based query logger

marschall:
Add a [JDK_Flight_Recorder](https://en.wikipedia.org/wiki/JDK_Flight_Recorder)(JFR) based alternative to `QueryStackTraceLogger`.

JFR can potentially be more efficient than logging, especially when multiple similar events are logged and UI tools for analysis exist.

The main disadvantage is it only works with HotSpot based JDKs and requires at least JDK 11.

dev.java has a good [introduction](https://dev.java/learn/jvm/jfr/) to JFR

A short summary:

- A JFR recording can be started using different ways
   - using [jcmd](https://docs.oracle.com/en/java/javase/21/docs/specs/man/jcmd.html)
   - using the `Recording` API
   - using the `-XX:StartFlightRecorder` JVM option
   - using JMC
 - A recording can be analyzed in different ways
   - using the `RecordingFile` API
   - using [JDK Mission Control](https://jdk.java.net/jmc/)(jmc)

Individual queries and then we viewed in a generic event view

QueryStatementLogger

or users can create dedicated views and for example group events by SQL string and JMC would then merge all events with the same stack trace

HibernateQueriesGrouped


vladmihalcea:
Another alternative is to use the [Vidual VM JDBC profiling](https://vladmihalcea.com/jdbc-profiling-visualvm/).

marschall:
Absolutely there are various options, each with with its own set of advantages and disadvantages. Consider this more a demo of JFR, I find it easer to show code to explain. No problem if you decide not to merge it.

vladmihalcea:
I will merge it, but I'll have to find some time to test it too.

vladmihalcea:
Fixed.

PR #659, opened 2023-09-22
Implement QueryStackTraceLogger using StackWalker

marschall:
Implement `QueryStackTraceLogger` using the [Stack-Walking API](https://openjdk.org/jeps/259).

This is potentially more efficient because:

- We render into one `StringBuilder`
  -   this avoids all intermediate, small `String` allocations
  - this also avoids the intermediate `List<String>` allocation holding the individual lines
- We only allocate `StackTraceElement` for the frames we are interested in
- We do not need to walk the frames after the first frame in `packageNamePrefix`
- We avoid allocating the `StackTraceElement[]`
- We do not walk the stack and create the log message if `QueryStackTraceLogger` has not at least level debug. This can be helpful if you want to always register `QueryStackTraceLogger` but be able to toggle it using the log level.

The code quite is still not as elegant as I would like. I especially hate the nested `if`s and the `AtomicBoolean` but the output is the same. It's also not ideal that the stack trace ends in `o.hypersistence.utils.hibernate.query.QueryStackTraceLoggerTest.lambda$0(QueryStackTraceLoggerTest.java:48)` rather than `io.hypersistence.utils.hibernate.query.QueryStackTraceLoggerTest.testJPQL(QueryStackTraceLoggerTest.java:39)`. There seems to be no way to determine whether a method is a lambda short of running a regex against the method name or checking `Method#isSynthetic()`.

This requires JDK 9. So you may only want to use is for Hibernate 6+ unless you want to build a [Multi-Release JAR](https://openjdk.org/jeps/238)


vladmihalcea:

vladmihalcea:
Fixed.

Issue #661, opened 2023-10-06
JPQL queries using LIKE are not working on MySQL because of the escaping character


runningonfumes:
As of version 3.4.4 the following error occurs when executing a like query.

JDBC exception executing SQL [select p1_0.id,p1_0.name from person p1_0 where p1_0.name like ? escape \]
You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '\' at line 1

The generated query contains escape \ which is not correct.

Before version 3.4.4 the escape character was provided as a parameter and the generated query was 

select p1_0.id,p1_0.name from person p1_0 where p1_0.name like ? escape ?

To reproduce, see attached zip file, this is a simple jpa test that will fail for 3.4.4 or later. If you update the pom to an earlier version the test will work.
[demo.zip](https://github.com/vladmihalcea/hypersistence-utils/files/12828548/demo.zip)

vladmihalcea:
This looks like a Hibernate issue. If you think it's a problem in a Type provided by this library, then you have to send a Pull Request with the fix.

I no longer have the time to investigate or fix issues for this library.

runningonfumes:
Ok, thank you. 

FYI, it seems this was already mentioned before in https://github.com/vladmihalcea/hypersistence-utils/issues/638
But the problem still exists in the latest release.

vladmihalcea:
That issue was fixed. If you bump into that with the latest version, then the first step would be to create a replicating test case that demonstrates the issue. Afterwards, you will need to create a fix for it.

runningonfumes:
I created a pullrequest : https://github.com/vladmihalcea/hypersistence-utils/pull/663

Is this the right approach?

vladmihalcea:
Closed as the Pull Request was not approved and no test case was provided.

PR #663, opened 2023-10-11
Mysql escape

runningonfumes:


vladmihalcea:

vladmihalcea:
@runningonfumes The tests provided don't match the LIKE query issue described in the #661 issue.

In fact, if you debug the tests, you will see that the overridden method is not even called, so the tests don't cover the code changes.

vladmihalcea:
@runningonfumes Also, it's worth investigating why the default `CharacterType` works and the `NullableType` does not for that query.

PR #671, opened 2023-11-02
Array query cache fix

victornoel:
Hi,

This is a (partial, see below) fix for broken query caching when using arrays in parameters.

For versions
- 5
  - I wasn't able to make the test run because of issues with instantiating the entity manager to check the hit/miss count on the cache
  - the fix is there and I expect it's working... ?
- 52, 55
  - the test and fix work well
  - I took the liberty of testing a few corner cases scenario when mixing primitive and boxed types
- 60
  - the test does not pass, even with the fix, because of a bug in hibernate
  - see the comment on the test's ignore annotation for more details
- 62
  - query cache was working out of the box, so I added a test
  - the test is simpler than with 52, 55 because hibernate throws exceptions when using boxed types 
  - I still added the fix in a separate commit for consistency with the `JavaType` interface

### DIFF
--- a/hypersistence-utils-hibernate-5/src/main/java/io/hypersistence/utils/hibernate/type/array/internal/AbstractArrayTypeDescriptor.java
+++ b/hypersistence-utils-hibernate-5/src/main/java/io/hypersistence/utils/hibernate/type/array/internal/AbstractArrayTypeDescriptor.java
@@ -66,6 +66,11 @@ public abstract class AbstractArrayTypeDescriptor<T>
     }
 
     @Override
+    public int extractHashCode(T value) {
+        return ArrayUtil.hashCode(value);
+    }
+
+    @Override
     public String toString(T value) {
         return Arrays.deepToString(ArrayUtil.wrapArray(value));
     }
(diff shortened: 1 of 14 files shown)


vladmihalcea:

vladmihalcea:
The test is not correct as it generates an HQL like this:

    SELECT id FROM Event WHERE sensor_names = ?1

The property is called `sesnorNames`, not `sensor_names`.

victornoel:
> The property is called `sensorNames`, not `sensor_names`.

Right, of course, I think Hibernate 6 became more strict about this because you can use `sensor_names` in Hibernate 5 it seems, that's strange but I agree, I should be using `sensorNames` all the time.

> Also, the tests are way too complicated by the Parameterized usage. If one of those fails, it's hard to debug it

Ok, as you prefer, I am used to parameterized tests but it's true the code is a bit complex as-is, I will replace them with individual tests for each combination that all calls some method to build the different test cases.

I will actually add both Criteria API and strings because I suspect the cache is involved differently depending on which you use, no? I don't see the problem with SQL injections, we are writing tests that we control and it's best to ensure caching works with both because end users DO use strings in their code.

victornoel:
Hello @vladmihalcea, in the end it seems I was too optimistic about being able to work on this. I got taken by other priorities and won't have the time to work on it. Sorry about that.

PR #673, opened 2023-11-07
resolves #668

### LINKED ISSUE #668: proxy breaks SQLExtractor
`SQLExtractor.from(TypedQuery)` didn't work from some of my use cases: for some reason, I got a proxy for my TypedQuery instance, which isn't assignable from QuerySqmImpl.

I could solve it with following workaround:

    private Query getHibernateQuery(TypedQuery<?> typedQuery) {
        try {
            if (typedQuery instanceof QuerySqmImpl) {
                return typedQuery;
            }
            // is proxyied, get it out
            InvocationHandler invocationHandler = Proxy.getInvocationHandler(typedQuery);
            Class<?> innerClass = invocationHandler.getClass();
            Field targetField = innerClass.getDeclaredField("target");
            targetField.setAccessible(true);
            return (Query) targetField.get(invocationHandler);
        } catch (NoSuchFieldException | IllegalAccessException exception) {
            throw new RuntimeException(exception);
        }
    }

I suggest including it in SQLExtractor.

ymajoros:


ymajoros:
https://github.com/vladmihalcea/hypersistence-utils/pull/675 actually solves it in a better way. They could still be split in 2, but I think it's not worth it, as further refactorings were needed.

vladmihalcea:
@ymajoros Why make all these changes instead of just doing this instead?

    String sql = SQLExtractor.from(criteriaQuery.unwrap(Query.class));

ymajoros:
@vladmihalcea this seems to indeed simplify it. I updated https://github.com/vladmihalcea/hypersistence-utils/pull/675 with that, as it was needed there as well.

PR #675, opened 2023-11-08
674 extract parameters

ymajoros:


vladmihalcea:

hypersistence-utils-hibernate-62/src/main/java/io/hypersistence/utils/hibernate/query/SQLExtractor.java
```diff
@@ -36,42 +45,96 @@ protected SQLExtractor() {
...
+     * @param querySqm the internal query SQM object
+     * @return a list of parameter values
+     */
+    private static List<Object> getParametersFromInternalQuerySqm(QuerySqmImpl<?> querySqm) {
+        List<Object> parameterValues = new ArrayList<>();
+
+        QueryParameterBindings parameterBindings = querySqm.getParameterBindings();
+        parameterBindings.visitBindings((queryParameterImplementor, queryParameterBinding) -> {
+            Object value = queryParameterBinding.getBindValue();
+            parameterValues.add(value);
```
KENNYSOFT, line 105:
```suggestion
            if (queryParameterBinding.isMultiValued()) {
                parameterValues.addAll(queryParameterBinding.getBindValues());
            } else {
                parameterValues.add(queryParameterBinding.getBindValue());
            }
```

If you call `getBindValue` to multi-valued binding, IllegalStateException("Binding is multi-valued; illegal call to #getBindValue") occurs.

Specifically, the extracted SQL of Hibernate 6.0+ using this class contains expanded JdbcParameters. For example, given HQL below:

```sql
select team
from Team team
where team.isDeleted = ?1 and team.no in ?2
```

turns into (abbreviated) SQL as:

```sql
select t1_0.no from team t1_0 where t1_0.is_deleted=? and t1_0.no in (?,?,?,?)
```

when IN clause parameter has 4 items.

KENNYSOFT (changes requested):
First of all, thanks, you saved my life!

I've added just one simple but important suggestion from my experience with these wonderful changes.
Also, you should copy these methods to hibernate-60 and hibernate-63 modules too.

I'm looking forward to this PR being merged, and released soon!

hypersistence-utils-hibernate-62/src/main/java/io/hypersistence/utils/hibernate/query/SQLExtractor.java
KENNYSOFT: Nah, if the IN clause parameter has 3 items (e.g., `[1, 2, 3]`) and it's padded to 4 items (e.g., `(?,?,?,?)`; by configuring something like `hibernate.query.in_clause_parameter_padding`), `bindValues` still have 3 items so the size does not match.

I've found the key from the original code! Here is the working code fragment (in Kotlin):

```kotlin
val parameterBindings = ReflectionUtils.invokeMethod<JdbcParameterBindings>(
    plan,
    ReflectionUtils.getMethod(
        plan,
        "createJdbcParameterBindings",
        ReflectionUtils.getClass<Any>("org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan\$CacheableSqmInterpretation"),
        DomainQueryExecutionContext::class.java,
    ),
    cacheableSqmInterpretation,
    querySqm,
)
return jdbcSelect.parameterBinders.map { parameterBindings.getBinding(it as JdbcParameter)?.bindValue }
```

Copy the `getSQLFromSqmQuery` method and replace the code in `if (jdbcSelect != null)` block to the fragment.

ymajoros: > Nah, if the IN clause parameter has 3 items (e.g., `[1, 2, 3]`) and it's padded to 4 items (e.g., `(?,?,?,?)`; by configuring something like `hibernate.query.in_clause_parameter_padding`), `bindValues` still have 3 items so the size does not match.
> 
> I've found the key from the original code! Here is the working code fragment:
> 
> ```java
> JdbcParameterBindings parameterBindings = ReflectionUtils.invokeMethod(
>         plan,
>         ReflectionUtils.getMethod(
>                 plan,
>                 "createJdbcParameterBindings",
>                 ReflectionUtils.getClass("org.hibernate.query.sqm.internal.ConcreteSqmSelectQueryPlan$CacheableSqmInterpretation"),
>                 DomainQueryExecutionContext.class
>         ),
>         cacheableSqmInterpretation,
>         querySqm
> );
> return jdbcSelect.getParameterBinders().stream().map(b -> {
>     JdbcParameterBinding parameterBinding = parameterBindings.getBinding((JdbcParameter) b);
>     if (parameterBinding != null){
>         return parameterBinding.getBindValue();
>     }
>     return null;
> }).toList();
> ```
> 
> Kotlin version
> Copy the `getSQLFromSqmQuery` method and replace the code in `if (jdbcSelect != null)` block to the fragment.

@KENNYSOFT I'm not working on this ATM, so if you rely on this functionality, I'd suggest you update the PR.

PR #681, opened 2023-12-03
Fix simple typo in README

kinow:


vladmihalcea:
Thanks for pointing it out. There were more issues, so I used Grammary to fix them all.

PR #683, opened 2023-12-05
fixes #682, allow negative durations

### LINKED ISSUE #682: Oracle Interval Type does not support negative intervals
**Describe the bug**
Oracle Interval Type cant handle negative duration like "PT-13H-5M". 

**To Reproduce**
Save duration "PT-13H-5M" -> error invalid interval.
Same issue when loading -> does not match regex with leading minus sign.

**Expected behavior**
Negative interval should be parsed to `INTERVAL '-0 13:05:00' DAY TO SECOND`  instead of `INTERVAL '0 -13:-05:00' DAY TO SECOND`

inola-ww:


vladmihalcea:
Thanks, I merged it.

PR #684, opened 2023-12-17
 Add hypersistence-utils-hibernate-63 module for Hibernate 6.3

### LINKED ISSUE #657: Add a hypersistence-utils-hibernate-63 module for Hibernate 6.3
Description:

An attempt was made to call a method that does not exist. The attempt was made from the following location:

`  io.hypersistence.utils.hibernate.type.basic.PostgreSQLEnumType.setParameterValues(PostgreSQLEnumType.java:56)`

The following method did not exist:

`org.hibernate.metamodel.model.convert.spi.EnumValueConverter org.hibernate.type.EnumType.getEnumValueConverter()`

tmvlpl:
These changes should make the code compatible with Hibernate 6.3 and fix #657.

**Details:**

- As suggested in the issue discussion I have removed the `PostgreSQLEnumType` and deprecated it in the other Hibernate 6 modules.
- The abstract base classes `MutableType` and `DescriptorImmutableType` now implement `DomainType<T>` instead of `SqmExpressible<T>`. It looks to me as though `DomainType` is a kind of `SqmExpressible` default implementation but I am a bit confused by all the different type interface hierarchies.
- The explicit cast in the `BindArrayTypeQueryParameterTest` seems to be necessary, as i could find no way to register the custom function return type with the hibernate sqm.
- I have updated the optional/test dependency versions in the new module  and  aligned them with the SpringBoot 3.2 BOM.


vladmihalcea:

YuriMB:
Does it make sense to support 6.3 since hibernate already dropped support? Maybe go straight to 6.4?

vladmihalcea:
@tmvlpl I'll merge this PR and create a new issue for 6.4.

vladmihalcea:
Fixed.

PR #704, opened 2024-02-29
Problem with JsonType serialization with spring-data 3.2 and hibernate 6.3

jongyoonlee:
I ran into this issue while migrating my app from spring boot 2 to spring boot 3. This PR contains a test case to illustrate the problem.

There are some Entity objects with Java collection type members that map to json column in a table. They are annotated with JsonType. I noticed the tests for persisting those entities using named query were failing after migrating to spring boot 3. It looks like the code path in hibernate for handling named query is completely different from code path for persisting. It ends up doing Java serialization instead of the custom serialization for JsonType.


vladmihalcea:
Try to debug it and see why it was working on Hibernate 5 and fails on Hibernate 6.

PR #706, opened 2024-03-13
Make JsonTypeDescriptor thread-safe

tony-tock:
We use this together with a library that initializes multiple hibernate persistence units concurrently (specifically, [here](https://github.com/tocktix/onami-persist/tree/6b0724c8ae00956a31c013479d2a2ee47efb17c8/src/main/java/org/apache/onami/persist/AllPersistenceUnits.java#L85-L99)) and recently started hitting an intermittent initialization error with a stack trace like this:

```java
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 0
	at java.util.ArrayList.add(ArrayList.java:455)
	at java.util.ArrayList.add(ArrayList.java:467)
	at io.hypersistence.utils.hibernate.type.json.internal.JsonTypeDescriptor.validatePropertyType(JsonTypeDescriptor.java:207)
	at io.hypersistence.utils.hibernate.type.json.internal.JsonTypeDescriptor.setPropertyClass(JsonTypeDescriptor.java:195)
	at io.hypersistence.utils.hibernate.type.json.internal.JsonTypeDescriptor.setParameterValues(JsonTypeDescriptor.java:82)
```

I believe the cause of this was a `static ArrayList` used in JsonTypeDescriptor. I'd like to change that to a thread-safe `Set` (there's a `contains` check that also needs to be consolidated to be thread-safe, and a `Set` seemed better suited) to prevent this.

I didn't think it would be valuable to attempt to reproduce the concurrency issue in a unit test, but I did add a test that exercise this code path. Happy to try and improve on that if you have any suggestions.

Thanks!

### DIFF
--- a/hypersistence-utils-hibernate-52/src/main/java/io/hypersistence/utils/hibernate/type/json/internal/JsonTypeDescriptor.java
+++ b/hypersistence-utils-hibernate-52/src/main/java/io/hypersistence/utils/hibernate/type/json/internal/JsonTypeDescriptor.java
@@ -201,10 +201,9 @@ public class JsonTypeDescriptor
                 ParameterizedType parameterizedType = (ParameterizedType) propertyType;
 
                 for(Class genericType : ReflectionUtils.getGenericTypes(parameterizedType)) {
-                    if(validatedTypes.contains(genericType)) {
+                    if(!validatedTypes.add(genericType)) {
                         continue;
                     }
-                    validatedTypes.add(genericType);
                     Method equalsMethod = ReflectionUtils.getMethodOrNull(genericType, "equals", Object.class);
                     Method hashCodeMethod = ReflectionUtils.getMethodOrNull(genericType, "hashCode");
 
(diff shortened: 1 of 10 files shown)


vladmihalcea:
@tony-tock That's a good catch, but I wonder how this issue would replicate with multiple `PersistenceUnit` instances since usually the bootstrapping is not done using multiple Threads?

tony-tock:
> @tony-tock That's a good catch, but I wonder how this issue would replicate with multiple `PersistenceUnit` instances since usually the bootstrapping is not done using multiple Threads?

@vladmihalcea I would imagine it's not a common setup, but we use a fork of an old Apache project called onami-persist where the persistence units are explicitly initialized across multiple threads:

https://github.com/tocktix/onami-persist/blob/6b0724c8ae00956a31c013479d2a2ee47efb17c8/src/main/java/org/apache/onami/persist/AllPersistenceUnits.java#L83-L100

vladmihalcea:
Fixed.

PR #716, opened 2024-05-08
Allow PostgreSQLHStoreType to process PGobject

Iapetus999:
Not sure if it's my jdbc driver or if I'm missing some annotation but I get `PGobject` in the result set.
This checks the type and does the appropriate conversion.


vladmihalcea:
I checked the [`PgResultSet`](https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/PgResultSet.java#L305), and the result is returned as `Map`.

So, to validate this change, we need a replicating test case.

vladmihalcea:
Closed as no replicating test case was provided

PR #721, opened 2024-05-27
Use Spring Data Range as an alternative to PostgreSQL range type

locrianmode:
Currently, to map the range Hibernate types, there are already 2 options

- Range class from Hypersistence Utils
- Range class from Guava

I would like to add Range class from the Spring Data as a third option.
This would facilitate using the same Range class for use with Spring Data JPA and Spring Data Elasticsearch and perhaps in other Spring Data projects.

The changes with tests are made only to the 6x modules, as the 5x modules uses the Spring Data commons 2.7x where the version of the Range class is incompatible when using Date/Time types e.g. LocalDate, ZonedDateTime etc.


vladmihalcea:
Thanks, I merged it.

PR #729, opened 2024-06-22
Add support for Batch IdGeneratorType

### LINKED ISSUE #728: Enhance BatchSequenceGenerator with support for IdGeneratorType
Hibernate 6.5 has deprecated `@GenericGenerator` see https://discourse.hibernate.org/t/deprecation-of-genericgenerator-in-hibernate-6-5/9482. Instead the `@IdGeneratorType` meta-annotation should be used https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#identifiers-generators-IdGeneratorType

The result is a lot less verbose, what previously was

```java
  @Id
  @GenericGenerator(
          name = "parent_id_generator",
          strategy = "com.github.marschall.hibernate.batchsequencegenerator.BatchSequenceGenerator",
          parameters = {
              @Parameter(name = SEQUENCE_PARAM, value = "SEQ_PARENT_ID"),
              @Parameter(name = FETCH_SIZE_PARAM, value = "50")
          })
  @GeneratedValue(generator = "parent_id_generator")
  private Long parentId;
```

Is now simply

```java
  @Id
  @BatchSequence(name = "SEQ_PARENT_ID", fetchSize = 50)
  private Long parentId;
```
Here is an example what the changes may look like  https://github.com/marschall/hibernate-batch-sequence-generator/commit/d88975a0be56703e4431d2bb829b47be6b409692

I'm not sure this can be retrofitted into the 6.3 module or a new 6.5 module is needed.

marschall:
Add a @BatchSequence meta annoation for BatchSequenceGenerator

Fixes #728

### DIFF
--- a/hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequence.java
+++ b/hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequence.java
@@ -1,0 +1,51 @@
+package io.hypersistence.utils.hibernate.id;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import org.hibernate.annotations.IdGeneratorType;
+
+/**
+ * Meta annotation to use {@link BatchSequenceGenerator} as an identifier generator.
+ * 
+ * @author Philippe Marschall
+ * @since 3.7.8
+ */
+@IdGeneratorType(BatchSequenceGenerator.class)
+@Retention(RUNTIME)
+@Target({ FIELD, METHOD })
+public @interface BatchSequence {
+
+    /**
+     * Returns the name of the sequence to use.
+     * 
+     * @return the name of the sequence to use
+     */
+    String name();
+
+    /**
+     *  Returns how many sequence values to fetch at once.
+     * 
+     * @return how many sequence values to fetch at once, must be positive
+     */
+    int fetchSize() default BatchSequenceGenerator.DEFAULT_FETCH_SIZE;
+
+    /**
+     * Returns the catalog name of the sequence to use.
+     * 
+     * @return the catalog name of the sequence to use
+     */
+    String catalog() default "";
+
+    /**
+     * Returns the catalog name of the sequence to use.
+     * 
+     * @return the catalog name of the sequence to use
+     */
+    String schema() default "";
+
+}
--- a/hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequenceGenerator.java
+++ b/hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequenceGenerator.java
@@ -2,6 +2,7 @@ package io.hypersistence.utils.hibernate.id;
 
 import org.hibernate.HibernateException;
 import org.hibernate.MappingException;
+import org.hibernate.annotations.GenericGenerator;
 import org.hibernate.boot.model.naming.Identifier;
 import org.hibernate.boot.model.relational.Database;
 import org.hibernate.boot.model.relational.QualifiedName;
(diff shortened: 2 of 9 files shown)


hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequence.java
```diff
@@ -0,0 +1,51 @@
...
+@Retention(RUNTIME)
+@Target({ FIELD, METHOD })
+public @interface BatchSequence {
+
+    /**
+     * Returns the name of the sequence to use.
+     * 
+     * @return the name of the sequence to use
+     */
+    String name();
```
marschall, line 28: An argument can be made this should be named `value` instead. That way the when going with the default fetch size the code is a little bit more concise

```java
@BatchSequence("SEQ_MY_ENTITY")
``` 
However when going with a custom fetch size it is a bit more awkward

```java
@BatchSequence(value = "SEQ_MY_ENTITY", fetchSize = 100)
``` 

let me know what you think.

hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequence.java
```diff
@@ -0,0 +1,51 @@
...
+     * @return how many sequence values to fetch at once, must be positive
+     */
+    int fetchSize() default BatchSequenceGenerator.DEFAULT_FETCH_SIZE;
+
+    /**
+     * Returns the catalog name of the sequence to use.
+     * 
+     * @return the catalog name of the sequence to use
+     */
+    String catalog() default "";
```
marschall, line 42: Unfortunately we can not use `null` as a default. We could also use a sentinel like `"<default>"`.

hypersistence-utils-hibernate-62/src/test/java/io/hypersistence/utils/hibernate/id/AbstractBatchSequenceGeneratorTest.java
```diff
@@ -93,15 +93,10 @@ protected long getSelectCount(QueryCount queryCount) {
...
 
         @Id
-        @GenericGenerator(
-            name = "post_sequence",
-            strategy = "io.hypersistence.utils.hibernate.id.BatchSequenceGenerator",
-            parameters = {
-                    @Parameter(name = "sequence", value = "SEQ_PARENT_ID"),
-                    @Parameter(name = "fetch_size", value = "" + BATCH_SIZE)
-            }
+        @BatchSequence(
```
marschall, line 96: I didn't manage to run the tests. `io.hypersistence.utils.spring.repo.base.SpringDataJPABaseTest` went to log level debug and they would never finish.

hypersistence-utils-hibernate-60/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequenceGenerator.java
```diff
@@ -141,12 +138,18 @@ public class BatchSequenceGenerator implements BulkInsertionCapableIdentifierGen
 
     /**
      * Indicates the name of the sequence to use, mandatory.
+     * 
+     * @deprecated use {@link BatchSequence}
      */
+    @Deprecated
```
marschall, line 144: We do not really have to deprecate these but since `GenericGenerator` is deprecated in 6.5 and `@BatchSequence` is already available I believe it is justified.

marschall (commented):
Comments for reviewers

marschall:
You're correct. It looks like we are running into [HHH-17878](https://hibernate.atlassian.net/browse/HHH-17878) which is only [fixed](https://github.com/hibernate/hibernate-orm/commit/ddb23247368599e0d8da5e90e8c1b2700889984d) in 6.5. The 6.3 tests only pass because they run with 6.5, they would fail with 6.3 and 6.4. I'll work on a fix.

marschall:
I think I found  a solution. I ran all three `H2BatchSequenceGeneratorTest` in my IDE and they now pass. I also updated the 6.3 submodule as I assume it is intended to work with 6.3 and 6.4 as well, not just 6.5.

vladmihalcea:
Thanks, I merged it.

PR #732, opened 2024-07-14
feat: Json CLOB support for Hibernate 6.x

agebhar1:
Using 

```java
@Type(JsonType::class)
@Column(columnDefinition = "clob")
```

fails because `java.sql.Blob` is not assignable from instance of `java.sql.Clob` in https://github.com/vladmihalcea/hypersistence-utils/blob/bbb27c4417d3809c9bfe0ca44e6b8f52710a2d87/hypersistence-utils-hibernate-63/src/main/java/io/hypersistence/utils/hibernate/type/json/internal/JsonJavaTypeDescriptor.java#L166-L194

Also `public <X> X unwrap(Object value, Class<X> type, WrapperOptions options) ` fails.


vladmihalcea:
Thanks, I merged it.

PR #745, opened 2024-10-24
Add support for SortedSet

markwoon:
Add support for SortedSet in ListArrayType.

Ideally it would pull comparator info from annotations, but I'm not sure how to get that from within the TypeDescriptor.


vladmihalcea:
Thanks, I merged it.

markwoon:
@vladmihalcea Is it possible to pull annotations from within the TypeDescriptor?

vladmihalcea:
@markwoon Check out [the `setParameterValues` from `JsonJavaTypeDescriptor` ](https://github.com/vladmihalcea/hypersistence-utils/blob/master/hypersistence-utils-hibernate-63/src/main/java/io/hypersistence/utils/hibernate/type/json/internal/JsonJavaTypeDescriptor.java#L76).

PR #759, opened 2024-12-28
Support timezone offsets in seconds for Range type (for historical dates in certain timezones)

sghoshjr:
Certain timezones have had offsets in seconds for historical dates, it seems. This causes the `Range<ZonedDateTime>` to throw errors while parsing these dates if the local system time is one of these timezones.

```sql
set timezone to 'Asia/Calcutta';
select tstzrange('1900-01-01 00:00:00Z', 'infinity');
--["1900-01-01 05:21:10+05:21:10",infinity)

set timezone to 'Europe/Paris';
select tstzrange('1900-01-01 00:00:00Z', 'infinity');
--["1900-01-01 00:09:21+00:09:21",infinity)
```

This PR provides a fix to support zonedDateTime with offsets in seconds.


vladmihalcea:
Thanks, I merged it.

PR #768, opened 2025-03-11
Correct resolve base type, when we use polymorphic collections, that not implement Serializable 

mvkirillov:


vladmihalcea:
I applied it after some modifications.

PR #803, opened 2025-08-16
Created a bunch of Converts. They are really handy while saving a list as a JSON in Database. 

sadiul-hakim:


vladmihalcea:
What is the purpose of this change? I need a detailed description to see what value doe sit bring to the project.

PR #812, opened 2025-10-08
Minor change to handle change in Hibernate 7.2.0.CR1

cigaly:
Here is minor change that fixes change in Hibernate 7.2.0.CR1


vladmihalcea:
Thanks, I merged it. I tested it with both 7.1.4 and 7.2.0.CR1 and works like a charm.

PR #815, opened 2025-11-27
Add null checker on xProperty MemberDetails on JsonJavaTypeDescriptor

jsolferreira:
Adds a null checker to prevent the issue https://github.com/vladmihalcea/hypersistence-utils/issues/814


vladmihalcea:
Thanks, I merged it.

jsolferreira:
Hi @vladmihalcea !
I'm afraid the code I pushed didn't actually fix the issue.
It seems that when the xProperty is null we need to get the returnedClass instead of the returnedJavaType, otherwise the property type is not resolved.

I've created this PR to invert the logic, but I'm afraid I don't know if this is the correct fix.

https://github.com/vladmihalcea/hypersistence-utils/pull/819

PR #825, opened 2026-01-26
Fix ObjectMapperJsonSerializer nested collection cloning

### LINKED ISSUE #808: ObjectMapperJsonSerializer fails to serialize types like `List<List<NonSerializableObject>>`
**Describe the bug**
In some scenarios entity attributes defined as follows
```java
        @Type(JsonBinaryType.class)
        @Column(columnDefinition = "jsonb")
        private List<List<NonSerializableObject>> any;
```
are serialized to `List<List<Map>>` and set back to entity instance so that when accessing the attribute will result something like this:

```
java.lang.ClassCastException: class java.util.LinkedHashMap cannot be cast to class NonSerializableObject
```

**To Reproduce**

After debugging I believe problem boils down to `ObjectMapperJsonSerializer::clone`. Here's the test case that fails and would resolve the issue if it passes.

```java
    @Test
    public void should_clone_matrix_with_non_serializable_nested_object() {
        Object original = List.of(
                List.of(new NonSerializableObject("name1")),
                List.of(new NonSerializableObject("name2"))
        );

        Object cloned = serializer.clone(original);

        assertEquals(original, cloned);
        assertNotSame(original, cloned);
    }
```
currently fails with

```
java.lang.AssertionError: 
Expected :[[name1], [name2]]
Actual   :[[{value=name1}], [{value=name2}]]
```

NineKoo9:
## Summary
Fix JSON clone behavior for nested collections so non-serializable leaf types are preserved during cloning.

Fixes #808

## Problem
`ObjectMapperJsonSerializer#clone` used `jsonClone` with only the raw collection class. 
For nested collections (e.g., `List<List<NonSerializableObject>>`), generic element types were lost at runtime, so Jackson deserialized inner JSON objects as `LinkedHashMap`. 
This caused equality checks and type assertions to fail (e.g., “Element should be NonSerializableObject but was LinkedHashMap”).

## Root Cause
Java type erasure + raw `Class` deserialization meant Jackson had no nested element type information.
The existing type inference only looked at the first-level element type and did not propagate nested generic type info.

## Fix
Resolve `JavaType` recursively for nested collections/maps and use it for JSON cloning.
This preserves leaf element types and prevents the `LinkedHashMap` fallback.

## Changes
- Preserve nested generic type information via recursive `JavaType` resolution.
- Use interface collection/map types when the runtime class cannot be instantiated safely.
- Consolidate clone test coverage into `ObjectMapperJsonSerializerTest`.

### DIFF
--- a/hypersistence-utils-hibernate-63/src/main/java/io/hypersistence/utils/hibernate/type/util/ObjectMapperJsonSerializer.java
+++ b/hypersistence-utils-hibernate-63/src/main/java/io/hypersistence/utils/hibernate/type/util/ObjectMapperJsonSerializer.java
@@ -7,10 +7,13 @@ import org.hibernate.internal.util.SerializationHelper;
 import org.hibernate.type.SerializationException;
 
 import java.io.Serializable;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.AbstractMap;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * @author Vlad Mihalcea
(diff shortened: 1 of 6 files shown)


NineKoo9:
I see your point now. It didn't cross my mind to simply enforce the requirement on JsonType.

Since JPA expects attributes to be serializable anyway, aligning with that makes sense. Thanks, I'll update it.

Issue #831, opened 2026-01-28
JsonType should not support Serializable attribute mappings


vladmihalcea:
The JPA specification requires that entity attributes be Serializable. Traditionally, this requirement was not enforced and led to various problems. The current solution for non-serializable types uses JSON object cloning, which is very slow.

Therefore, the default `JsonSerializer` will use the Java Serialization mechanism, and if you want to use non-serializable entity attributes, then you can provide your own `JsonSerializer`, as explained in [this article](https://vladmihalcea.com/how-to-customize-the-json-serializer-used-by-hibernate-types/)

lthoulon-locala:
Hi, I wanted to share some feedback on this change, even though I understand the fix itself is straightforward.

I agree with @StefanLobbenmeierObjego that this constitutes a breaking change. We use Renovate for automated dependency updates, and since this was released as a patch version, we didn't anticipate needing code changes. This means we now need to manually review and update each affected project individually.

In my case, I fortunately discovered this during a manual upgrade on a single project (we were still on _62, so Renovate hadn't auto-updated to _63 yet). However, this could have been more disruptive if it had rolled out automatically across our projects.

I completely understand that the change is minimal and improves the library overall. I'm simply flagging this as something that had an impact on our workflow.

For context, the previous implementation was working well for us—we weren't experiencing noticeable latencies, at least none that triggered our monitoring or alerting systems.

Thanks for maintaining this excellent library!

radistao:
@vladmihalcea it seems after this new implementation the old tests became a bit misleading, https://github.com/vladmihalcea/hypersistence-utils/blob/master/hypersistence-utils-hibernate-71/src/test/java/io/hypersistence/utils/hibernate/type/util/ObjectMapperJsonSerializerTest.java#L56-L57 :

* ```java
  try {
    serializer.clone(original);
    fail("Should throw exception");
  } catch (Exception expected) {}
  ```
  this will success in either case, because _any_ exception (either from `clone()` or `fail()`) is silently caught and swallowed
  * i could imagine smth like `assertEquals(NonSerializableObjectException.class, expected.getClass());` above was expected

* all names <code>should_clone_***non_serializable_object</code> actually should become <code>should_<b>throw</b>\_when_clone_***non_serializable_object</code>

vladmihalcea:
Those tests should be removed, indeed.

PR #833, opened 2026-01-28
Improve the non-serializable error message in the ObjectMapperJsonSerializer to indicate the culprit Java class

CodeSimcoe:
Related to #831 

If the text block refactor is not relevant, tell me and I'll drop it.
Looked more readable to me.


CodeSimcoe:
Hello @vladmihalcea , i have restricted the change to its minimal form : pass the cause to the `IllegalArgumentException`.
It is the useful change, it allowed me to have detailed information about offending types and fix them all, in my case.
Please consider it.

PR #837, opened 2026-02-24
Added hypersistence-utils-hibernate-72 for Hibernate 7.2 with Jackson 3 support

dickerpulli:
I just reused the hypersistence-utils-hibernate-73 and back ported it to Hibernate 7.2.

Usefull for all having dependencies to Hibernate 7.2 and also Jackson 3, like Spring Boot 4.0.x has.

See comment on issue https://github.com/vladmihalcea/hypersistence-utils/issues/810#issuecomment-3951321334


dickerpulli:
Hibernate 7.2 has dependency to Jackson 2

PR #840, opened 2026-03-16
EnumArrayType: Persist enums by name() rather than by toString() as on read enum's valueOf() is used, which looks up Enums by name

noffke:
When using `EnumArrayType` with an enum where `name()` and `toString()` differ, the enum can't be persisted or read (depending on the definition) anymore, as `EnumArrayType` persists via `toString()` but reads via enum's `valueOf()` which uses `name()`.


vladmihalcea:
Reverted as tests were failing:

```
[ERROR] Errors:
[ERROR]   MultiDimensionalArrayTypeTest.test:57->AbstractHibernateTest.doInJPA:496 
```

PR #851, opened 2026-06-02
bump versions hibernate + fix hibernate 7.4.0.Final

oscarfanchin:
Hi @vladmihalcea , thanks for taking a look at this PR.

This change updates the Hibernate versions across the affected modules and adds a fallback option for running the tests against Hibernate 7.3 with module hypersistence-utils-hibernate-73 when needed. It also includes the small type-related adjustment required for Hibernate 7.4 compatibility.

- updated hypersistence-utils-hibernate-63 from Hibernate 6.6.38.Final to 6.6.52.Final
- updated hypersistence-utils-hibernate-71 from Hibernate 7.2.0.Final to 7.2.17.Final
- updated hypersistence-utils-hibernate-73 from Hibernate 7.3.0.CR1 to 7.4.0.Final
- added a fallback profile for tests targeting Hibernate 7.3.7.Final in the hypersistence-utils-hibernate-73 module
   the fallback can be enabled with -Dhibernate73
- adjusted MutableDynamicParameterizedType for compatibility with Hibernate 7.4

### DIFF
--- a/hypersistence-utils-hibernate-73/pom.xml
+++ b/hypersistence-utils-hibernate-73/pom.xml
@@ -14,6 +14,20 @@
 
     <name>hypersistence-utils-hibernate-73</name>
     <description>Utilities for Spring and Hibernate ORM 7.3</description>
+    
+    <profiles>
+	    <profile>
+	        <id>hibernate-73</id>
+	        <activation>
+	            <property>
+	                <name>hibernate73</name>
+	            </property>
+	        </activation>
+	        <properties>
+	            <hibernate.version>7.3.7.Final</hibernate.version>
+	        </properties>
+	    </profile>
+	</profiles>
 
     <dependencies>
 
@@ -217,7 +231,7 @@
         <maven.compiler.release>${jdk.version}</maven.compiler.release>
         <maven.compiler.testRelease>${jdk-test.version}</maven.compiler.testRelease>
 
-        <hibernate.version>7.3.0.CR1</hibernate.version>
+        <hibernate.version>7.4.0.Final</hibernate.version>
 
         <jackson-module.version>3.0.0</jackson-module.version>
         <guava.version>32.1.3-jre</guava.version>
--- a/hypersistence-utils-hibernate-73/pom.xml
+++ b/hypersistence-utils-hibernate-73/pom.xml
@@ -15,20 +15,6 @@
     <name>hypersistence-utils-hibernate-73</name>
     <description>Utilities for Spring and Hibernate ORM 7.3</description>
     
-    <profiles>
-	    <profile>
-	        <id>hibernate-73</id>
-	        <activation>
-	            <property>
-	                <name>hibernate73</name>
-	            </property>
-	        </activation>
-	        <properties>
-	            <hibernate.version>7.3.7.Final</hibernate.version>
-	        </properties>
-	    </profile>
-	</profiles>
-
     <dependencies>
 
         <dependency>
--- a/hypersistence-utils-hibernate-73/src/main/java/io/hypersistence/utils/hibernate/type/MutableDynamicParameterizedType.java
+++ b/hypersistence-utils-hibernate-73/src/main/java/io/hypersistence/utils/hibernate/type/MutableDynamicParameterizedType.java
@@ -6,13 +6,8 @@ import java.sql.CallableStatement;
 import java.sql.SQLException;
 import java.util.Properties;
 
-import org.checkerframework.checker.initialization.qual.Initialized;
-import org.checkerframework.checker.nullness.qual.NonNull;
-import org.checkerframework.checker.nullness.qual.Nullable;
-import org.checkerframework.checker.nullness.qual.UnknownKeyFor;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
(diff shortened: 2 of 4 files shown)


.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,2 @@
+eclipse.preferences.version=1
```
vladmihalcea: Those should never be committed since they are Eclipse-related

hypersistence-utils-hibernate-63/.gitignore
```diff
@@ -0,0 +1,3 @@
+/.project
```
vladmihalcea: Those two are not needed. You should have them in your local fork only.

hypersistence-utils-hibernate-73/src/main/java/io/hypersistence/utils/hibernate/type/MutableDynamicParameterizedType.java
```diff
@@ -1,23 +1,30 @@
 package io.hypersistence.utils.hibernate.type;
 
-import io.hypersistence.utils.hibernate.type.util.Configuration;
+import static jakarta.persistence.metamodel.Type.PersistenceType.BASIC;
+
+import java.sql.CallableStatement;
+import java.sql.SQLException;
+import java.util.Properties;
+
+import org.checkerframework.checker.initialization.qual.Initialized;
```
vladmihalcea: Why do we need an extra library? You can do everything with JUnit.

hypersistence-utils-hibernate-73/src/test/java/io/hypersistence/utils/hibernate/AAAHibernateVersionTest.java
```diff
@@ -0,0 +1,10 @@
+package io.hypersistence.utils.hibernate;
+import org.junit.Test;
+
+public class AAAHibernateVersionTest {
```
vladmihalcea: A class called AAA? No, thanks.

hypersistence-utils-hibernate-73/src/test/java/io/hypersistence/utils/hibernate/AAAHibernateVersionTest.java
```diff
@@ -0,0 +1,10 @@
...
+
+    @Test
+    public void printVersion() {
+        System.out.println("Using hibernate: " + org.hibernate.Version.getVersionString());
```
vladmihalcea: Console logging? Why?

hypersistence-utils-hibernate-73/pom.xml
```diff
@@ -14,6 +14,20 @@
 
     <name>hypersistence-utils-hibernate-73</name>
     <description>Utilities for Spring and Hibernate ORM 7.3</description>
+    
+    <profiles>
```
vladmihalcea: Don't create extra profiles that you're the only one using.

vladmihalcea:
Is this backward-compatible with Hibernate 7.3? If it's not, it will require a new module, apart from removing everything that's not needed for the fix, as explained in the comments above.

oscarfanchin:
They ended up there by mistake, I’m cleaning it up. Give me a second.

hypersistence-utils-hibernate-73/src/test/java/io/hypersistence/utils/hibernate/AAAHibernateVersionTest.java
oscarfanchin: Removed, is my local class that evidence version of hibernate

hypersistence-utils-hibernate-73/src/test/java/io/hypersistence/utils/hibernate/AAAHibernateVersionTest.java
oscarfanchin: Removed, is my local class that evidence version of hibernate

hypersistence-utils-hibernate-73/pom.xml
oscarfanchin: I added that profile only to verify that the change could still be tested against Hibernate 7.3 without constantly editing the POM, but I’ve removed it from the PR.
From my side, I did not see regressions with Hibernate 7.3.

.settings/org.eclipse.core.resources.prefs
oscarfanchin: Removed

hypersistence-utils-hibernate-73/src/main/java/io/hypersistence/utils/hibernate/type/MutableDynamicParameterizedType.java
oscarfanchin: You’re right, that extra library was not needed, and I’ve removed those references.

oscarfanchin:
The profile I added — and then removed as requested — was only there to make it easier to test the change against Hibernate 7.4.0.Final and then rerun the test suite against the latest Hibernate 7.3.7 (last version of 7.3).

Based on those test runs, I did not see regressions with Hibernate 7.3.

If you still prefer handling this in a separate module, I can rework the PR accordingly.

PR #855, opened 2026-06-04
#854 Use CONNECT BY LEVEL instead of CONNECT BY ROWNUM

### LINKED ISSUE #854: On Oracle 23/26, BatchSequenceGenerator fails with 'ORA-30009: Not enough memory for CONNECT BY operation'
**Describe the bug**
BatchSequenceGenerator can fail on Oracle 23/26 with `ORA-30009: Not enough memory for CONNECT BY
operation` exception.

**Additional context**
BatchSequenceGenerator is invoking `SELECT sequence.NEXTVAL FROM dual CONNECT by rownum <= ?` which works as long as the Oracle optimizer chooses proper plan.
Up until Oracle 19, it seemed to be always the case and the statement yielded correct results.

However changes in optimizer of Oracle 23/26 may lead to situation when it chooses different execution plan (with COUNT STOPKEY capping the CONNECT BY WITHOUT FILTERING) and in that case the statement fails with ORA-30009 exception.

The problem lies with the usage of `rownum` and Oracle and other sources recommend to use `level` instead.

> Why ROWNUM is the problem (and LEVEL is not):
> In a hierarchical query over dual (one row, no START WITH, no PRIOR):
> 
> CONNECT BY LEVEL <= N — LEVEL is the depth of the current node. Each node has exactly one child (the single dual row) as long as LEVEL <= N, so you get a linear chain of exactly N rows. Deterministic.
> CONNECT BY ROWNUM <= N — ROWNUM is a running count of output rows, evaluated independently of tree depth. Whether/when it prunes the walk depends entirely on the plan. When the engine doesn't push the stop condition into the walk, the expansion is no longer bounded the way you expect, and the buffered intermediate set explodes — hence the "Not enough memory for CONNECT BY operation."
> 

**Proposed fix**
Simple: invoke `SELECT sequence.NEXTVAL FROM dual CONNECT by level <= ?` instead of `SELECT sequence.NEXTVAL FROM dual CONNECT by rownum <= ?` for Oracle deatabases

mletenay:
Fix the Oracle case of BatchSequenceGenerator, by using LEVEL instead of ROWNUM.

The SELECT sequence.NEXTVAL FROM dual CONNECT by rownum <= ? was working more or less by accident on older versions of Oracle and no longer reliably works on Oracle 26.
It works if optimizer produces execution plan with COUNT STOPKEY capping the CONNECT BY WITHOUT FILTERING (which is the case of e.g. Oracle <23). However if the optimizer chooses different plan the query with rownum will run off and fails with "ORA-30009: Not enough memory for CONNECT BY operation".

Fixes #854

### DIFF
--- a/hypersistence-utils-hibernate-63/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequenceGenerator.java
+++ b/hypersistence-utils-hibernate-63/src/main/java/io/hypersistence/utils/hibernate/id/BatchSequenceGenerator.java
@@ -88,7 +88,7 @@ import java.util.concurrent.locks.ReentrantLock;
  * <pre><code>
  * SELECT seq_xxx.nextval
  * FROM dual
- * CONNECT BY rownum &lt;= ?
+ * CONNECT BY level &lt;= ?
  * </code></pre>
  *
  * <h3>SQL Server</h3>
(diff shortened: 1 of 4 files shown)


marschall:
LGTM, recommend merging

vladmihalcea:
Applied upstream, thanks

PR #859, opened 2026-07-14
Add PostgreSQL macaddr type support

### LINKED ISSUE #858: Add support for the PostgreSQL macaddr type
It looks like there is no support for Postgres macaddr type. Is there a specific reason for this, or is it an open area where a contribution would be welcome

Chiraggupta0:
Closes #858 

There wasn't a mapping for the PostgreSQL `macaddr` column type, so I added one by following how the existing `inet` type is done.

What I added (in the hibernate-63, 71 and 73 modules):
- `MacAddress` – a small value class that holds the address string
- `PostgreSQLMacAddressType` – an `ImmutableType<MacAddress>` that reads/writes the column using a PGobject of type "macaddr"
- registered the new type in `HibernateTypesContributor` next to `PostgreSQLInetType`

I named the type `PostgreSQLMacAddressType` to follow the Hibernate naming conventions, as suggested in the issue.

Testing:
- Added `PostgreSQLMacAddressTypeTest`, based on the existing `PostgreSQLInetTypeTest`
- It covers find, JPQL, native query and a raw JDBC check
- All tests pass on Testcontainers PostgreSQL in all three modules


vladmihalcea:
Thanks, I merged it.

PR #864, opened 2026-08-25
fix(array): resolve enum array type for Envers audit entities

### LINKED ISSUE #863: EnumArrayType throws NullPointerException when the entity is audited with Hibernate Envers
**Bug Description**
An Entity uses EnumArrayType for a mapping of a Java array of enum values to a Postgres array of corresponding Postgres enum values. When the entity is audited with Envers, an NPE is thrown when trying to access a  revision of the entity (not the revision number, but the actual revision entity).

Storing and loading the entity the regular way is OK. 
_Storing_ the envers audit is also ok.
Just _loading_ the revision entity fails.

Observations from debugging and stepping:

- Hypersistence has EnumArrayType and EnumArrayTypeDescriptor.
- EnumArrayTypeDescriptor does the transformation from Java to Postgres and vice versa
- This is done in method unwrap(), which calls into ArrayUtils.
- The descriptor needs the Java type for the Enum Array
- When dealing with entities directly, this is OK, the Java type is set.
- But the descriptor for the audit table (XXX_AUD) does not have this Java type set! Then unwrap() fails.
- It seems to be a difference between getParameterType vs getParameterClass. See notes in the workaround. 

**To Reproduce**
Step 1: add org.hibernate.orm:hibernate-envers as dependency 
Step 2: Use PostgreSQLEnumArrayTypeTest as a base and do the following modifications.
Copy the original entity UserAccount into UserAccountAudited,
give the entity a different name (UserAccountAudited) and add the Annotation Audited from hibernate envers.
For clarity, I have also initialized the array roles with UserRole[0], as there were hints that it is not good to have the value = null.

```
@Entity(name="UserAccountAudited") // <== different name
    @Table(name="users")
    @Audited // <== annotation for Envers Auditing
    public static class UserAccountAudited {
        @Id
        @GeneratedValue
        private Long id;
        private String username;
        @Type(
                value = EnumArrayType.class,
                parameters = @org.hibernate.annotations.Parameter(
                        name = "sql_array_type",
                        value = "user_role"))
        @Column(
                name = "roles",
                columnDefinition = "user_role[]")
        private UserRole[] roles = new UserRole[0]; 

        public Long getId() {
            return id;
        }
        public void setId(Long id) {
            this.id = id;
        }
        public String getUsername() {
            return username;
        }
        public void setUsername(String username) {
            this.username = username;
        }
        public UserRole[] getRoles() {
            return roles;
        }
        public void setRoles(UserRole[] roles) {
            this.roles = roles;
        }
    }
   
    @Override
    protected Class<?>[] entities() {
        return new Class<?>[] {
                UserAccount.class,
                UserAccountAudited.class // <== add the entity definition
        };
    }

// Additional test:
// 1. Create the entity and store to trigger the creation of the revision and the audit record.
// 2. Load the entity to verify that the entity itself is OK
// 3. Load the revision numbers and verify that the entity has indeed be audited.
// 4. Load the revision for the entity - here the NPE happens

   @Test
    void test_audited() {
        UserRole[] userRoles = {UserRole.ROLE_ADMIN, UserRole.ROLE_USER};
        doInJPA(entityManager -> {
            UserAccountAudited account = new UserAccountAudited();
            account.setUsername("vladmihalcea.com");
            account.setRoles(userRoles);
            entityManager.persist(account);
        });

        doInJPA(entityManager -> { 
            UserAccountAudited singleResult = entityManager.createQuery(
                            """
                            select ua
                            from UserAccountAudited  ua
                            where ua.username = :username
                            """,
                            UserAccountAudited.class)
                    .setParameter("username", "vladmihalcea.com")
                    .getSingleResult();
            assertThat(singleResult).as("vladmihalcea.com").isNotNull();
            assertThat(singleResult.getRoles()).as("Roles").isEqualTo(userRoles);

            AuditReader auditReader = AuditReaderFactory.get(entityManager);
            List<Number> revisions = auditReader.getRevisions(UserAccountAudited.class, singleResult.getId());
            // one Revision (ADD)
            assertThat(revisions).as("Revisions").isNotNull().isNotEmpty().hasSize(1);

            Number rev = revisions.get(0);
            Long id = singleResult.getId();
// the following line fails with an NPE:
            UserAccountAudited revisioned = auditReader.find(UserAccountAudited.class, id, rev);

        });
    }
```

**Expected behavior**
The entity revision can be loaded and contains the correct data.

**Additional context**

Our workaround for each type of Enum for which we would need EnumArrayType:

- Enum is **OurEnum**
- Define **OurEnumArrayType** extends **AbstractArrayType**<OurEnum[]>
- Define **OurEnumArrayTypeDescriptor** extends **AbstractArrayTypeDescriptor**<OurEnum[]>
- make sure that after the constructor has been called that the Enum Array Type is set!

OurEnumArrayType:
```
public class OurEnumArrayType extends AbstractArrayType<OurEnum[]> {
    private static final String DEFAULT_TYPE_NAME = "%s_enum_array_type";
    private String name;
    public OurEnumArrayType() {
        super(new OurEnumArrayTypeDescriptor(OurEnum[].class));
    }

    public OurEnumArrayType(Configuration configuration) {
        super(new OurEnumArrayTypeDescriptor(OurEnum[].class), configuration);
    }

    publicOurEnumArrayType(Class arrayClass, String sqlArrayType) {
        super(new OurEnumArrayTypeDescriptor(arrayClass));
        Properties parameters = new Properties();
        parameters.setProperty(SQL_ARRAY_TYPE, sqlArrayType);
        parameters.put(DynamicParameterizedType.PARAMETER_TYPE, new ParameterizedParameterType(arrayClass));
        setParameterValues(parameters);
    }

    public OurEnumArrayType(org.hibernate.type.spi.TypeBootstrapContext typeBootstrapContext) {
        this(new Configuration(typeBootstrapContext.getConfigurationSettings()));
    }

    public String getName() {
        return name;
    }

    @Override
    public void setParameterValues(Properties parameters) {
        ParameterType parameterType = (ParameterType) parameters.get(DynamicParameterizedType.PARAMETER_TYPE);
        if (name == null) {
            name = String.format(DEFAULT_TYPE_NAME, parameters.getProperty(SQL_ARRAY_TYPE));
        } 
// the following calls eventually into the descriptor:
        super.setParameterValues(parameters);
    }
}
```
OurEnumArrayTypeDescriptor:
```
public class OurEnumArrayTypeDescriptor extends AbstractArrayTypeDescriptor<OurEnum[]> {
   private String sqlArrayType;

    public OurEnumArrayTypeDescriptor() {
        super(OurEnum[].class);
    }

    public OurEnumArrayTypeDescriptor(Class enumClass) {
        super(enumClass);
    }

    @Override
    protected String getSqlArrayType() {
        return sqlArrayType;
    }

    @Override
    public void setParameterValues(Properties parameters) {
        // Notes
        // parameters.get(PARAMETER_TYPE) returns a ParameterType.
        // On getReturnedJavaType(), this parameter type returns null, but we need
        // den OurEnum[].class (Array of Enums).

        // The original Hypersistence Code uses the deprecated ParameterType,
        // and calls getReturnedType(), which returns null instead of a type.
        // However: getReturnedClass() returns the correct class for the Java side, 
        // here it would be OurEnum[].class.
        // *Type* vs *Class* !

// the following block is elaborate to illustrate the modifications for the original code: 
        if (getArrayObjectClass() == null && parameters.containsKey(PARAMETER_TYPE)) {
            Class<?> arrayObjectClassFromParms = ((ParameterType) parameters.get(PARAMETER_TYPE)).getReturnedClass(); // <== getReturned Class not Type!
            setArrayObjectClass(OurEnum[].class);
        }

        sqlArrayType = parameters.getProperty(AbstractArrayType.SQL_ARRAY_TYPE);
    }
}
```

arimu1:
## Summary

- Fixes NPE when loading Envers revision entities that use `EnumArrayType` on PostgreSQL enum arrays (#863).
- `AbstractArrayTypeDescriptor.setParameterValues` now falls back to `ParameterType.getReturnedClass()` when `getReturnedJavaType()` is null — the pattern Envers uses for audit-table mappings.
- Adds a focused unit test reproducing the Envers parameterization contract and a PostgreSQL + Envers integration test (runs in CI with Testcontainers).

## Root cause

Envers parameterizes the audit-table `EnumArrayType` with a `ParameterType` where `getReturnedJavaType()` is `null` but `getReturnedClass()` returns the concrete enum array type (e.g. `UserRole[].class`). The previous code assigned `null` to `arrayObjectClass`, causing `ArrayUtil.unwrapArray` to NPE when loading a revision via `AuditReader.find()`.

## Test plan

- [x] `EnumArrayTypeDescriptorTest.setParameterValuesUsesReturnedClassWhenReturnedJavaTypeIsNull` — fails on HEAD, passes with fix (proves scout)
- [x] `PostgreSQLEnumArrayTypeAuditedTest.testAuditedRevisionCanBeLoaded` — persist audited entity, load revision via `AuditReader.find()`, assert roles round-trip (requires Docker/Testcontainers; not run locally)
- [ ] CI green on `hypersistence-utils-hibernate-73`

Fixes #863


LissySes (approved):
Thanks for the fast fix!!

LissySes:
We have tested with Hypersistence Utils 73, Release 3.16.0 and the test fail is now fixed. Thanks.

vladmihalcea:
@LissySes You're welcome.

